PR #14, opened 2015-04-29
Highlight the 'volatile' keyword.

### LINKED ISSUE #11: No volatile keyword
In java there is a keyword "volatile", it works very similar to syncronized, but only in variables.

yihangho:
Fix #11.

:tada:

Screenshot


kevinsawicki:
Looks great, thanks for adding this :ship:

PR #16, opened 2015-07-29
Tokenize 'this' for consistency

### LINKED ISSUE #6: Update "this" to add a ".this" class
See: atom/language-coffee-script#39

Per the discussion on the CoffeeScript language package, it makes sense to keep the various default Atom language packages consistent. So rather than `this` being given a class of `variable.language` it should be given a class of `variable.language.this`.

Victorystick:
See: atom/language-coffee-script#39

Fixes #6

Added minimal specs which were completely absent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: objective-c
+
+notifications:
+  email:
+    on_success: never
+    on_failure: change
+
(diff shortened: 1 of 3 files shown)


50Wliu:
:green_heart: :fireworks:

PR #21, opened 2015-10-01
Fix overeager enum tokenization

### LINKED ISSUE #17: Methods not highlighted properly after constructor
Have to put the constructor after the method for it to be properly highlighted.
brokenjavahighlighting

50Wliu:
Previously, methods such as `A(int a, int b) { }` were being incorrectly tokenized as enums.

Fixes #17

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -331,6 +331,9 @@
         'include': '#comments'
       }
       {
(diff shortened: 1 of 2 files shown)


kevinsawicki:
:+1:

PR #22, opened 2015-10-01
Tokenize punctuation (and generics)

### LINKED ISSUE #19: Missing scopes on brackets and parentheses
Brackets and parentheses aren't scoped, they are the same color as variables:

javab
```java
list[i] = (sum == n ? true : false);
```

Sometimes they can be distinguished in an identifier class: 

javag
```java

}
```

But they still aren't scoped:

javac

### LINKED ISSUE #23: Missing scopes for angle brackets and commas
Angle brackets and commas aren't tokenized when used for generic type parameters:

java1

html1

java3

html3

### LINKED ISSUE #29: Implementing Type with Type parameter as Array highlights weird
implementing a Type with an array as the type parameter does not highlight correctly (e.g., "class A implements B\<C\[\]\> " -- shown below)

This looks strange

50Wliu:
This PR tokenizes a lot of the missing Java punctuation, such as parentheses, square and curly brackets, angle brackets and commas (for generics), etc.

Fixes #19
Fixes #23
Fixes #29


chbk:
Looks great, except the square brackets aren't tokenized everywhere. Maybe that's what you mean with `char[]`?

This is my (quick) test case:

javabraces

```java
public class MyClass {
    
        String[] array;
        
        HashMap<Integer,String> map = new HashMap<>();
        
        boolean[] list = new boolean[10];
        
        list[i] = (sum == n ? true : false);
    }
}
```

Brackets on line 3 aren't scoped:

line3

But on line 5 they are:

line5

I also noticed bugs on lines 7 and 9 while typing this, i'll open new issues for that.

50Wliu:
Yeah, that's basically what I was talking about.  Though it looks like they sometimes get tokenized, so I'll try to fix that :).

kevinsawicki:
Just a heads up, this no longer merges now that #21 has been merged.

50Wliu:
@chbk your example seems to pass now.  I have one question though: can you please take a look at [this](https://github.com/atom/language-java/blob/ebfe99ff2a2d9adaa0e48875857072546ad9078e/grammars/java.cson#L880) pattern and see if brackets also need to be matched in that?  I can't figure out what it's supposed to match at all.

chbk:
I'm not sure about that pattern. I'm guessing it tries to find variable names in bulk, so this would pass
`int abc[], test = 9;`
I don't think you should bother with brackets there, especially if they are already tokenized by the "code" section.

50Wliu:
Updates!

I finally got back around to working on this.  This should 100% fix #19 now and specs are passing :green_apple: which is a great start, but I still want to include #23 before merging.

Sorry for the delay @chbk ;(.

chbk:
Thanks for taking the time to work on this @50Wliu

50Wliu:
#23 should be fixed as well now :tada:!

@chbk could you test this out and report if there's still any places that aren't getting tokenized correctly?  Thanks! :bow:

grammars/java.cson
```diff
@@ -694,20 +724,7 @@
   'object-types-inherited':
     'patterns': [
       {
-        'begin': '\\b((?:[a-z]\\w*\\.)*[A-Z]+\\w*)<'
-        'end': '>|[^\\w\\s,<]'
-        'name': 'entity.other.inherited-class.java'
```
50Wliu: Unfortunately I couldn't find a way to keep this `name` with the `include`.  @kevinsawicki do you know if this should be possible, or was it intentionally not allowed?

kevinsawicki: > do you know if this should be possible, or was it intentionally not allowed?

At first glance, it does appear this would have to be removed to get the new generics patterns working. Nothing jumps out to my as to why this scope would be helpful so I think it is okay to remove it.

50Wliu:
The latest commit a67ef17 changes all commas to be `punctuation.separator.delimiter`.  Not sure if that's the best naming for it though: /cc @atom/languages (first time I'm using this, yay!).

50Wliu:
Just realized that this doesn't tokenize the comma in all scenarios yet, such as `int a, b = 3`.

chbk:
Here is everything I could think of:

t

```java
public class MyClass<K, V extends A & B & C> {                                        // Ampersands

    private class CodeMap<S extends Object, T> extends HashMap<S,T> {}

    public <T> void copy(HashMap<String, ? super T> src, Integer[] values) {
        
        String[] array;

        boolean[] list = new boolean[10];
        
        int i, sum, n;                                                                // Commas

        list[i] = (sum == n ? true : false);
        
        CodeMap<String, ? extends ArrayList> codebook;                                // Everything in < >
        codebook = new CodeMap<String,Code>(10,20,30);                                // Commas in ( )
        
        codebook . put("Konami",                                                      // Period
                        new Code('U', 'U', 'D', 'D', 'L', 'R', 'L', 'R', 'B', 'A')    // Commas
                      );
    }
}
```

50Wliu:
@chbk The period in your example specifically isn't matched right now: `(?<=\\S)\\.(?=\\S)`.  Only periods without surrounding whitespace are included.  Should that be changed?

Also, what should the `&`s be tokenized as?

chbk:
> Only periods without surrounding whitespace are included. Should that be changed?

Yes, adding spaces or line breaks is still valid java syntax.

> Also, what should the &s be tokenized as?

The docs on this are still in progress (https://github.com/atom/docs/pull/114) so I'd say `punctuation.separator.types` but @MaximSokolov would know better.

MaximSokolov:
> `punctuation.separator.types`

:+1:

50Wliu:
Everything in the code snippet provided has been addressed except the `CodeMap<String, ? extends ArrayList> codebook;` example.  That one will be the hardest to fix.

50Wliu:
Took me more than a couple of tries, but I finally got that last example working!  @chbk feel free to take another go at trying to break this.

chbk:
Syntax highlighting looks much better now, great job!
Regarding #29, I compiled and ran some tests and it works fine. I also found another bug:

f

```java
class MyClass {
    
    public Map<Integer,String> method() {}  // Missing scopes
    
}
```

It seems return types are pretty inconsistently scoped in general:

```java
class MyClass {

    // `String` is just meta.method.return-type
    public String returnString() {}

    // `boolean` is meta.method.return-type > storage.type.primitive.array
    public boolean returnBoolean() {}
    
}
```

50Wliu:
@chbk Fixed, fixed, and fixed :smile:.

Your `Map` example was something that I caused by accident in 1bf118e when moving `object-types` around.  I've fixed that (hopefully) by placing it in `all-types`.

Same thing with `String` - objects weren't getting matched in return types.
`boolean` being tagged as an array was a silly bug.  `[]` is now required to call something an array, instead of being optional...

chbk:
It works as expected now. I combed through my files and found another live one:

t

```java
abstract class Document implements Comparable,Serializable {}       // Comma isn't tokenized
```

50Wliu:
When in doubt, keep fixing.

50Wliu:
@chbk just wanted to ping you in case you didn't see my last comment and commit :wink:.

chbk:
Sorry I didn't get back to you, your last commit works fine :+1:
Unfortunately I haven't been working with Java lately, so I haven't encountered any other bugs for the moment.

chbk:
I got back to this and found 2 other bugs:

p

```java
package com.apple.quicktime.v2;     // Periods aren't scoped
```

b

```java
class Test {
        
    public String[][] return2DStringArray() {    // Brackets
        
        String[][] array = new String[2][2];
        return array;
    }
    
    public int[][] return2DintArray() {          // Brackets
        
        int[][] array = new int[2][2];
        return array;
    }
}
```

50Wliu:
I think I can do a quick-fix for the brackets but the real underlying issue is #24.

50Wliu:
Apparently I was wrong...multidimensional arrays turned out to be a lot easier than expected to implement.

Think there's anything else to tackle @chbk?

chbk:
All these changes are starting to add up indeed! I don't see anything else, besides #24 and perhaps these minor enhancements:

- The package wildcard could use a scope: `import mypackage.*;`

- `instanceof` is just `keyword.operator`, making it hard to differentiate from others.

50Wliu:
@chbk hopefully this will be the last test!

grammars/java.cson
```diff
@@ -6,71 +6,96 @@
...
+      }
+      {
         'match': '\\.'
         'name': 'punctuation.separator.java'
       }
       {
-        'match': '\\s'
-        'name': 'invalid.illegal.character_not_allowed_here.java'
+        'match': '\\*'
+        'name': 'punctuation.wildcard.java'
```
50Wliu: @MaximSokolov Do you think this naming is ok?

MaximSokolov: Wouldn't `variable.language.wildcard` be better? I don't think it's a punctuation mark.
Note how syntax themes would highlight it:
> screen shot 2016-02-03 at 11 05 44
DuoTone dark

MaximSokolov:
How about wrapping brackets with `meta.brace`:
```coffee
{
  'begin': '(\\[)'
  'beginCaptures':
    '0':
      'name': 'meta.brace.square.scope-is-for-back-capability.java'
    '1':
      'name': 'punctuation.bracket.square.java'
  # ...
}
```
This allows to use `punctuation` scope and won't break themes which style `meta.brace` (e.g. Solarized)

grammars/java.cson
```diff
@@ -448,10 +492,56 @@
         'captures':
           '1':
             'name': 'keyword.operator.dereference.java'
```
MaximSokolov: Also how about changing it to `punctuation.separator.period` I think period are less important than other operators:
> screen shot 2016-02-04 at 10 32 44
Monokai

50Wliu: What about `punctuation.separator.deference`?  Or are we going with labeling what the actual character is now?

MaximSokolov: `punctuation.separator.accessor ` maybe?

50Wliu:
I think this is ready for review now.

50Wliu:
:ship:ing to get some exposure.  I'm fairly confident this works as expected, and there's extensive spec coverage.

PR #26, opened 2015-10-31
Add Java 8 Arrow (->) to keywords/operators

i-tu:
Currently the arrow is separated into the minus and greater-than
operators. Combining them as a single operator is more correct, and is
required for the Hasklig font arrow to render correctly

### DIFF
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -68,3 +68,15 @@ describe 'Java grammar', ->
     expect(lines[2][10]).toEqual value: ')', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.identifier.java']
     expect(lines[3][1]).toEqual value: '{', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.begin.java']
     expect(lines[4][1]).toEqual value: '}', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.end.java']
+
+  it 'tokenizes arrow operator', ->
+    lines = grammar.tokenizeLines '''
+      (String s1) -> s1.length() - outer.length();
+    '''
+
+    expect(lines[0][1]).toEqual value: 'String', scopes: ['source.java', 'storage.type.java']
+    expect(lines[0][5]).toEqual value: '->', scopes: ['source.java', 'keyword.operator.arrow.java']
+    expect(lines[0][7]).toEqual value: '.', scopes: ['source.java', 'keyword.operator.dereference.java']
+    expect(lines[0][9]).toEqual value: '(', scopes: ['source.java', 'meta.method-call.java', 'punctuation.definition.method-parameters.begin.java']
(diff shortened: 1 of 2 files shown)


50Wliu:
Please add specs to this.

i-tu:
Sure!

I hope these are sufficient. Let me know if otherwise.

spec/java-spec.coffee
```diff
@@ -68,3 +68,15 @@ describe 'Java grammar', ->
     expect(lines[2][10]).toEqual value: ')', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.identifier.java']
     expect(lines[3][1]).toEqual value: '{', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.begin.java']
     expect(lines[4][1]).toEqual value: '}', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.end.java']
+
+  it 'tokenizes arrow operator', ->
+    lines = grammar.tokenizeLines '''
```
50Wliu: Useful tip: this can be `tokenizeLine` instead (see line 16 for an example).  Then `lines[0][x]` can be simplified to `tokens[x]`.

i-tu:
I checked the [Java 8 spec](https://docs.oracle.com/javase/specs/jls/se8/jls8.pdf), and it's slightly ambigous about what  `->` is called.

On page 721, the exact name seems to be left out:
>*LambdaExpression:*
>*LambdaParameters* -> *LambdaBody*

Sect. 15.2 (page 466) indirectly calls `->` the "arrow" of a lambda expression.
> The lowest precedence operator is the arrow of a lambda expression (->), followed
by the assignment operators.

As far as I can tell, either "lambda operator" or "arrow operator" would be correct. Based on some googling, both are widely used. I'll leave it up to you.

MaximSokolov:
According to [naming conventions](http://manual.macromates.com/en/language_grammars#naming_conventions) it's `storage.type.function.arrow`

i-tu:
Let me know if this change to the class also requires moving the `->` regexp to another place in the grammar file.

50Wliu:
> Let me know if this change to the class also requires moving the -> regexp to another place in the grammar file.

Yeah, that would definitely be appreciated.

i-tu:
Sorry for the back and forth. First time contributing to any kind of language grammar.

Note: Matching for `(->)` is done purposefully before matching for
`(\\-|\\+|\\*|\\/|%)` (*keyword.operator.arithmetic.java*). The aforementioned regexp could be amended to `(\\-(?!>)|\\+|\\*|\\/|%)` so lambda expressions are never matched to minus symbols. Similarly, `>` regexp could be amended so that when they contain a preceding `-`, they are never mapped to the "greater than" sign.

I suppose it is much more simple to precede by the lambda expression regexp.

50Wliu:
> I suppose it is much more simple to precede by the lambda expression regexp.

Yep!

MaximSokolov:
@50Wliu :ship:

50Wliu:
:boat:

i-tu:
🎉

PR #30, opened 2016-02-02
Rewrite `new` regex patterns

### LINKED ISSUE #24: Primitive types not scoped consistently
The first `boolean` is "primitive" as it should be, but the second one isn't:

java2

html2

### LINKED ISSUE #32: Inline classes not detected if bracket is on another line
For example:

``` java
class AClass {
  public void someMethod() {
    // compare
    method( new Foo()
      {
        void bar( int a, int b ) {
          // ...
        }
      });

    // with
    method( new Foo() {
      void bar( int a, int b ) {
        // ...
      }
    });
  }
}
```

50Wliu:
Use what we're given instead of trying to reinvent everything.

Depends on #22
Fixes #24
Fixes #32


chbk:
Looks like a bug here:

```java
boolean[] list = new boolean[variable];
```

`variable` is `storage.type.primitive.array.java`

50Wliu:
Fixed!

chbk:
Ok primitives work nice but there's another bug:

f

```java
public class MyClass {

    public String insert(String key, String value) {
        return map.put(key, new Value(value));          // Last parenthesis
    }                                                   // Brace
}                                                       // Brace
```

The punctuation is `meta.method-call.java`.
Sorry for the late replies, I'm coming back to this as I can.

50Wliu:
@chbk Thanks for catching that; I would never have noticed that because I don't give any special styling to brackets (and that's precisely why you're filing all these punctuation issues and not me...).  In the latest commit I also fixed calls to constructors being identified as a storage type and gave the method call a better name (`entity.name.function.java` instead of `meta.method-call.java`).

chbk:
@50Wliu I'm glad to help if I can. I checked again with the code in issue #32 and I'm seeing a bug:

```java
class AClass {
  public void someMethod() {
    method( new Foo()
    {

    });     // parenthesis
  }         // brace
}           // brace
```

Weirdly enough, if I insert a whitespace after `Foo()`, the parenthesis gets scoped correctly.

50Wliu:
This is going to be trickier than I thought...my previous fix doesn't cover all the scenarios either.

For example,
```java
map.put(key,
        new Value(value)
); 
```
still fails.

50Wliu:
@chbk so it looks like in the time between my last comment and now, significant improvements have been made to language-java such that all your examples now seem to work.  If you wouldn't mind testing again...:smile:.

PR #37, opened 2016-05-16
Correctly match primitive types

### LINKED ISSUE #36: Any primitive type gets 'storage.type.primitive.array.java' even if not array
Reproduction:

```java
class Test {
  public int a;
}
```

put cursor at `int` and `Editor: Log Cursor Scope`:
screen shot 2016-05-16 at 6 05 35 pm

(adding `[]` after int as to actually have a primitive array works as expected)

noseglid:
A array of a primitive type may not be followed by a word
boundary. For example

```java
int[] myArray = new int[42];
```

Is perfectly valid.

So is also

```java
int     [] myArray;
```

so zero or more whitespaces after `int`. `\s*` should
be the right separator.

Fixes #36

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -862,7 +862,7 @@
   'primitive-arrays':
     'patterns': [
       {
-        'begin': '\\b(void|boolean|byte|char|short|int|float|long|double)\\b\\s*(?=\\[)'
+        'begin': '\\b(void|boolean|byte|char|short|int|float|long|double)\\s*(?=\\[)'
         'end': '(?!\\s*\\[)'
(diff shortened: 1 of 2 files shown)


50Wliu:
Can you add a spec for this please?

noseglid:
Sure thing, done.

50Wliu:
I'm having a bit of trouble understanding this fix - why is `public int hi` getting tagged as an array when there's an explicit lookahead for `[`, and how does this fix the issue?

noseglid:
I'm uncertain. I don't seem to be able to reproduce the case now either. I'll close this and get back to you if I manage to reproduce it.

PR #39, opened 2016-05-23
Correctly differentiate between variables and methods

noseglid:
A variable always ends with with a semi-color or equal sign.
Check for this or method definitions may be confused for
variables.


noseglid:
This seemed to only be a problem when there was a storage modifier (I am not sure why...). That's why the other test didn't catch it.

50Wliu:
Can I see a before/after please?

noseglid:
Sure thing cap'n!

Without fix
without fix

With fix
with fix

50Wliu:
:ship:

PR #40, opened 2016-05-24
Do not set scope "return-type" to space after return type

noseglid:
The space after the storage type in a method return value should not be tagged with `meta.method.return-type.java`.

Before
screen shot 2016-05-24 at 4 07 39 pm

After
screen shot 2016-05-24 at 4 06 51 pm


50Wliu:
Can you add an explicit test for this to make sure the space is scoped properly?

PR #41, opened 2016-05-24
Just because a variable is capitalized doesn't mean it's constant

noseglid:
Having this rule here means it did not get the scope
`meta.definition.variable.name.java` which is what we're sure it is.

Possibly the scope `constant.variable.java` should be added if
there is a storage modifier `final`. I'll leave that for another PR.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -966,12 +966,6 @@
           {
             'captures':
               '1':
-                'name': 'constant.variable.java'
-            'match': '([A-Z_0-9]+)\\s+(?=\\=)'
-          }
-          {
-            'captures':
-              '1':
                 'name': 'meta.definition.variable.name.java'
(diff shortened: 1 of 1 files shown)


50Wliu:
I'd like to leave this in for now since the majority of the time ALL_CAPS variables are constants.

> Possibly the scope constant.variable.java should be added if
> there is a storage modifier final.

This sounds like a much better solution :+1:.

noseglid:
I understand your reasoning, but I still think this PR corrects a faulty behavior. If `ALL_CAPS` are finals or not really has nothing to do with the name of the variable. I've been working on projects where variables like that existed.

Variables like `_1` and `_A` will also be constants.

50Wliu:
Ok, that sounds reasonable.

PR #42, opened 2016-05-26
Tokenize instance variables correctly

50Wliu:
Fixes:
* comments not being recognized
* storage modifier not being identified as part of the variable declaration when using primitive types or `def`

Also changed the capture name from `meta.definition.variable.name.java` to `variable.definition.java`.


noseglid:
Why are you removing the token on the name of the defined variable?

50Wliu:
I didn't remove it, I changed it from `meta.definition.variable.name` to `variable.definition` as `meta` isn't supposed to be used there.

noseglid:
Ah, right you are. This looks good to me! 👍

50Wliu:
Thanks @noseglid!

Issue #33, opened 2016-05-04
Dollar should be part of identifiers


PhiLhoSoft:
I opened a similar issue for JavaScript: https://github.com/atom/language-javascript/issues/374

Frankly, I dont use Atom to edit Java code, I use Eclipse instead. Given the low number of issues here, either I am not alone, or you are very good... :stuck_out_tongue: 

OK, let's repeat the gist of the previous issue:

Atom 1.7.3, Windows 7, although the latter is probably not relevant...

AFAIK, $ is part of the valid characters in identifiers, like underscore. Yet it is treated as a foreign punctuation character in many features of Atom.
Examples:
- Double click to select the identifier: the $ isn't selected (if prefix or suffix) or stops the selection (in the middle).
- Word move with Ctrl+Left / Right arrow: stops on $
- Search **log** with Whole Word option: selects the _log_ part of `$log`.
- If an identifier has a $ in its middle, the part after the $ is highlighted differently.

I know that Sun recommended not to use $ in identifiers, keeping this usage for themselves, but yet it is a valid char, and should be treated as such.

noseglid:
Variables with `$` is correctly highlighted now.

Correct me if I'm wrong, but I don't think the "double-click" or `ctrl-left` to jump words is managed by the languages. @50Wliu ?
If not than this can be closed.

PhiLhoSoft:
> Correct me if I'm wrong, but I don't think the "double-click" or ctrl-left to jump words is managed by the languages

If not, it is a problem (with Atom itself) / they should be: these operations are based on the concept of "words", and this concept is tightly related to the language: $ is a good example here (valid in JS / Java, not in other languages), but in other languages (CSS for example) dash is also part of identifiers (words), and so on.

noseglid:
@PhiLhoSoft I totally agree. It should be part of the language definition. But as it is now, the rest of this issue (aside from highlighting) does not belong to this repo (*if* that is so).

In CSS, stuff like dashes in e.g. `border-top` behave the same way as how you're describing `$` in java.

PhiLhoSoft:
That's annoying, but indeed, if that's beyond your scope, please feel free to close the issue if you fixed what was within your reach.

50Wliu:
> Correct me if I'm wrong, but I don't think the "double-click" or ctrl-left to jump words is managed by the languages. @50Wliu ?

[...]

PR #43, opened 2016-05-27
Differentiate between function and method calls

50Wliu:
`meta.method.java` has been renamed to `entity.name.function.java`.  This rename in particular should add a lot more "highlighting" to Java files.

/cc @noseglid

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -416,6 +416,9 @@
         'include': '#method-call'
       }
       {
+        'include': '#function-call'
+      }
+      {
         'include': '#strings'
       }
       {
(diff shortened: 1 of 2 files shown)


noseglid:
One thing that may not be matched:
```
this.method
    (arg1, arg2);
```

Perhaps also a spec for the `$` thingie, it's to fix #33 ?

50Wliu:
That is unfortunately something that can't be tokenized correctly yet due to the multiline restriction documented  at atom/first-mate#57.

noseglid:
I see. And the multiline, i cannot be solved with `beginMatch` and `endMatch`?
I'm still learning about the textmate grammars. I'm sorry if the answers to my questions are obvious.

This looks good now. I gave it a whirl and couldn't find any issues.

50Wliu:
`begin` and `end` only solve issues where the containing content can be on multiple lines, eg for a function, but `begin`/`end` can also only span one line.  That can't be used here because the parentheses that indicates that `this.method` is a method and not a property is on a new line.

PR #44, opened 2016-05-29
Capitalized variables are not (always) constants.

noseglid:
Same reasoning as in #41


noseglid:
Before
screen shot 2016-05-29 at 8 28 28 pm

After
screen shot 2016-05-29 at 8 28 54 pm

PR #45, opened 2016-05-30
Tokenize objects and properties

50Wliu:
This PR adds support for objects and properties, and is based off of language-javascript PRs by @MaximSokolov (as was the last PR that added function/method call support).

This PR also fixes a bug where objects were being tokenized as storage types, such as `System` in `System.out.println()`.

A side-effect of these changes is drastically different syntax highlighting, which caused some protest in language-javascript when those changes reached stable.

/cc @noseglid once again (but this should be the last one for a while)

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -419,6 +419,12 @@
         'include': '#function-call'
       }
       {
+        'include': '#objects'
+      }
+      {
+        'include': '#properties'
+      }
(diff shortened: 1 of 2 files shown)


noseglid:
Can't see anything wrong with this. My syntax theme doesn't seem to react differently for this.

LGTM 👍

Issue #50, opened 2016-06-10
Array variable declarations are not highlighted


50Wliu:
![array-no-highlighting](https://cloud.githubusercontent.com/assets/2766036/15965870/5a1e0906-2eee-11e6-9f79-a333a0091a29.png)

```java
class A
{
    private int array[];
}
```

noseglid:
I can give this a go. Been meaning to rewrite the variable structure.

I suppose your example is missing storage type?
```java
class A
{
    private int array[];
}
```

Regarding the variable structure, I want to split it into two separate categories. One (called something like `member-fields`) will match storage modifers and such (`private`, `protected`, `static`, etc). It will then include the `variables` structure (which match variables as they are matched today, but without modifiers).

50Wliu:
> I suppose your example is missing storage type?

Oops, fixed! :see_no_evil: 

Your new structure sounds like a good idea.

Issue #47, opened 2016-06-08
Qualified instance variables are not highlighted


50Wliu:
EDIT: see https://github.com/atom/language-java/issues/47#issuecomment-264060903
![qualified-no-highlight](https://cloud.githubusercontent.com/assets/2766036/15911740/0cc1e1e6-2d9f-11e6-82a7-308f685d8ad9.png)
vs
![variable-highlighting](https://cloud.githubusercontent.com/assets/2766036/15911816/68e66c44-2d9f-11e6-8409-2acb60151aa8.png)

``` java
class A
{
    private Test.Double something; // Double is tagged as `variable.other.property.java`
    private Double something2;
}
```

noseglid:
I also noted that generics are not scoped properly unless they have a visibility modifier. Could it be the same issue?
![screen shot 2016-06-09 at 8 30 53 am](https://cloud.githubusercontent.com/assets/933880/15920724/93cde338-2e1c-11e6-8db5-1a0b1f016040.png)

50Wliu:
Possibly!

50Wliu:
@noseglid this actually just got worse with #53:
![even-worse](https://cloud.githubusercontent.com/assets/2766036/15994720/ea98be26-30d9-11e6-8184-41029dc69bfb.png)

50Wliu:
Something else:
![partial-highlighting](https://cloud.githubusercontent.com/assets/2766036/15994813/8cfb29e0-30dc-11e6-8ecf-6359f9634d0e.png)

noseglid:
Both comments addresses in #58

50Wliu:
This is getting better.
![qualified-better-highlighting](https://cloud.githubusercontent.com/assets/2766036/20779589/70783018-b743-11e6-88e0-e348e81916fc.png)
`Double` just needs to be tokenized properly now.

PR #53, opened 2016-06-10
New variables structure

noseglid:
Do not include storage modifiers in the variables matches
because it's not great to re-use them in code blocks in that
case.

This also made it possible to put `#variables` as an
include in `#code`, which removes the need for having
both `#code` and `#variables` in many blocks.

One difference here is that storage modifiers (`private`, `final`,
`static`, etc) will no longer be tagged with `meta.definition.variable.java`,
and it shouldn't be IMO. It's not part of the definition, but rather a modification to
a definition. Syntax highlighting remains the same.

This also fixes my comment in #47, but not #47 itself.

Fixes: #50


grammars/java.cson
```diff
@@ -377,19 +391,7 @@
...
-        'end': '}'
-        'endCaptures':
-          '0':
-            'name': 'punctuation.section.block.end.bracket.curly.java'
-        'patterns': [
-          {
-            'include': '#code'
-          }
-        ]
+        'include': '#anonymous-block-and-instance-initializer'
```
noseglid: No actual changes, just some splitting up code to make it more readable.

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
       }
     ]
   'variables':
-    'applyEndPatternLast': 1
+    'begin': '''
+      (?x:(?=
+        (?:
```
50Wliu: Since we're in a lookahead all the non-capturing groups aren't needed.

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
...
-              |
-              (?:(?:[a-z]\\w*\\s*\\.\\s*)*[A-Z]+\\w*) # object type
-            )
-            \\s+
-            (?!private|protected|public|native|synchronized|abstract|threadsafe|transient|static|final)
-            [\\w\\d_<>\\[\\],\\?][\\w\\d_<>\\[\\],\\?\\s]*
-            \\s*(=|;)
-          ))
-        '''
+        'match': '([A-Za-z$_][\\w$]*)(?=(\\[\\])?\\s*(?=;|=|,))'
```
50Wliu: There can be spaces before the array brackets, eg `int test [] = new int[3];`.

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
...
+        'captures':
+          '1':
+            'name': 'variable.definition.java'
```
50Wliu: I realized after I merged the PR adding this that the scope should probably be `variable.other.definition.java` instead.

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
...
-              }
-            ]
-          }
-          {
             'include': '#code'
           }
         ]
       }
+      {
+        'include': '#code'
```
50Wliu: Why is this needed?

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
...
+      }
     ]
+    'end': '(?=;)'
+    'name': 'meta.definition.variable.java'
+  'member-variables':
+      'begin': '(?=private|protected|public|native|synchronized|abstract|threadsafe|transient|static|final)'
```
50Wliu: Your indentation is off here

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
...
             'include': '#code'
           }
         ]
       }
+      {
+        'include': '#code'
+      }
     ]
+    'end': '(?=;)'
+    'name': 'meta.definition.variable.java'
```
50Wliu: As in the other PR, end and name before patterns.

spec/java-spec.coffee
```diff
@@ -60,9 +60,10 @@ describe 'Java grammar', ->
   it 'tokenizes punctuation', ->
     {tokens} = grammar.tokenizeLine 'int a, b, c;'
 
-    expect(tokens[2]).toEqual value: ',', scopes: ['source.java', 'punctuation.separator.delimiter.java']
-    expect(tokens[4]).toEqual value: ',', scopes: ['source.java', 'punctuation.separator.delimiter.java']
-    expect(tokens[6]).toEqual value: ';', scopes: ['source.java', 'punctuation.terminator.java']
+    console.log tokens
```
50Wliu: :scream:

grammars/java.cson
noseglid: Right you are. Copied this from the previous definition, but might as well simplify!

grammars/java.cson
noseglid: 👍

grammars/java.cson
noseglid: Yes, that seems to be the consesus in textmate. Changing.

grammars/java.cson
noseglid: To get e.g. punctuation in `int a, b, c;`

grammars/java.cson
noseglid: 👍

grammars/java.cson
noseglid: Nice catch. Seems to work all the same. I hate CoffeeScript/cson for this reason. "Indentation is dead important! sometimes..."

spec/java-spec.coffee
noseglid: ouch. sloppy :( I blame late night coding

grammars/java.cson
```diff
@@ -999,50 +1001,54 @@
       }
     ]
   'variables':
-    'applyEndPatternLast': 1
```
noseglid: I removed this.  To be frank, I don't know how It worked with it before, since `;` is matched by `#code` which was included before too.

50Wliu: Surprisingly first-mate actually recognizes this.  The super-sparse documentation seems to suggest that this only matters when the end patterns of a subpattern and the overarching `end` are exactly the same, but if everything already works, then no need to have it!

noseglid: Cool. In my manual tests it seems to work well, and the specs agrees.

noseglid:
NB. Added this to commit message
> One difference here is that storage modifiers (`private`, `final`,
`static`, etc) will no longer be tagged with `meta.definition.variable.java`,
which is correct IMO. It's not a definition, but rather a modification to
a definition. Syntax highlighting remains the same.

50Wliu:
:boat:

50Wliu:
A multi-array variable still isn't highlighted correctly, eg `private int stats[][];`.

noseglid:
screen shot 2016-06-13 at 9 01 35 am

Looks okay to me... what do you mean?

Issue #54, opened 2016-06-11
Support highlighting for javadoc


sadikovi:
Currently, as of version 0.17.0 of the package, there is no javadoc highlight. I could not find option to enable it either. What do you think about adding this feature? Or have you considered adding it in the past? For example, `language-scala` supports highlighting, so we could just implement something very similar, I guess.

@50Wliu I would appreciate your feedback on this, thanks.

50Wliu:
I'm definitely :+1: on adding javadoc support.

sadikovi:
PR has been merged, I will close it.

PR #55, opened 2016-06-12
Add javadoc comments

sadikovi:
This PR adds javadoc comments to `language-java`: `@author`, `@deprecated`, `@exception`, `@return`, `@see`, `@serial`, `@since`, `@throws`, `@version`, `@param`, and `@link`, mentioned in issue #54. I tried adding highlighting for embedded code snippet in javadoc comments, but could not make it work.

Used [language-scala](https://github.com/atom-community/language-scala) package mainly for guidance and implementation, since Atom packages are new to me.

Here is what it looks like in Atom:
sample

#### Testing
Updated existing `tokenizes enum` test, added unit-tests.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -507,7 +507,6 @@
         '0':
           'name': 'punctuation.definition.comment.java'
       'name': 'comment.block.javadoc.java'
-      'contentName': 'meta.comment.block.javadoc.java'
       'patterns': [
         {
           'match': '@(author|deprecated|return|see|serial|since|version)\\b'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -462,6 +462,46 @@
   'comments-inline':
     'patterns': [
       {
+        'begin': '^\\s*/\\*\\*'
+        'captures':
+          '0':
+            'name': 'punctuation.definition.comment.java'
+        'end': '\\*/'
+        'name': 'comment.block.java'
+        'patterns': [
+          {
+            'captures':
+              '1':
+                'name': 'keyword.other.documentation.javadoc.java'
+              '2':
+                'name': 'variable.parameter.java'
+            'match': '(@param)\\s+(\\S+)'
+          }
+          {
+            'captures':
+              '1':
+                'name': 'keyword.other.documentation.javadoc.java'
+              '2':
+                'name': 'entity.name.type.class.java'
+            'match': '(@(?:exception|throws))\\s+(\\S+)'
+          }
+          {
+            'captures':
+              '1':
+                'name': 'keyword.other.documentation.javadoc.java'
+              '2':
+                'name': 'entity.name.type.class.java'
+              '3':
+                'name': 'variable.parameter.java'
+            'match': '\\{(@link)\\s+(\\S+)?#([^\\{\\}]+)\\}'
+          }
+          {
+            'match': '@(author|deprecated|exception|return|see|serial|since|throws|version)\\b'
+            'name': 'keyword.other.documentation.javadoc.java'
+          }
+        ]
+      }
+      {
         'begin': '/\\*'
         'captures':
           '0':
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -493,7 +493,7 @@
                 'name': 'entity.name.type.class.java'
               '3':
                 'name': 'variable.parameter.java'
-            'match': '\\{(@link)\\s+(\\S+)?#([^\\{\\}]+)\\}'
+            'match': '\\{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]+\\)).*\\}'
           }
           {
             'match': '@(author|deprecated|exception|return|see|serial|since|throws|version)\\b'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -534,7 +534,7 @@
               'name': 'entity.name.type.class.java'
         }
         {
-          'match': '\\{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]*\\)).*\\}'
+          'match': '{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]*\\)).*}'
           'captures':
             '1':
               'name': 'keyword.other.documentation.javadoc.java'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -100,6 +100,9 @@
     ]
   }
   {
+    'include': '#comments-javadoc'
+  }
+  {
     'include': '#code'
   }
 ]
@@ -335,6 +338,9 @@
   'class-body':
     'patterns': [
       {
+        'include': '#comments-javadoc'
+      }
+      {
         'include': '#comments'
       }
       {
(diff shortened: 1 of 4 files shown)


sadikovi:
@50Wliu @noseglid please review.

grammars/java.cson
```diff
@@ -462,6 +462,46 @@
   'comments-inline':
     'patterns': [
       {
+        'begin': '^\\s*/\\*\\*'
+        'captures':
+          '0':
+            'name': 'punctuation.definition.comment.java'
+        'end': '\\*/'
+        'name': 'comment.block.java'
```
noseglid: Possibly have `contentName: meta.comment.body.java` for consistency with class bodies, method bodies, try-catch-finally bodies etc.

noseglid:
Javadoc comments should only be in the class (enum,interface etc) body, right? As it is now it also tags javadoc such as:
```java
class Test {
    /**
     * @throws somethingElse
     */
  }
}
```

Which may not be what we want.

grammars/java.cson
```diff
@@ -462,6 +462,46 @@
...
+          }
+          {
+            'captures':
+              '1':
+                'name': 'keyword.other.documentation.javadoc.java'
+              '2':
+                'name': 'entity.name.type.class.java'
+              '3':
+                'name': 'variable.parameter.java'
+            'match': '\\{(@link)\\s+(\\S+)?#([^\\{\\}]+)\\}'
```
noseglid: [After `#something` there can be a `label`](http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIECH). This will be tagged with the `variable.parameter.java` which is probably not correct. Adding that to the specs would be nice too

grammars/java.cson
sadikovi: I guess I should replace `'name': 'comment.block.java'` with `'contentName': 'meta.comment.body.java'` in the whole `comments-inline`, right?

sadikovi:
@noseglid Yeah, you are right about javadoc inside method, this "feature" seems to also exist in `language-scala`:). I am not sure how to fix it, to be honest. Do you know, if  I can extract information if javadoc is inside method body somehow?

sadikovi:
Well, I will think about how to possibly fix `javadoc-inside-method-body` issue.

grammars/java.cson
sadikovi: I mean only for line 470 and line 510.

noseglid: I believe the name is correct (the tags `/**` and `*/` will get `name`). If you add `contentName` then the *content* will get that name - which is the body of the comment. What do you think, @50Wliu ?

noseglid:
I think I would create a `javadoc` pattern, and only include it in the `class-body` or the "root" (for javadocing classes).

grammars/java.cson
50Wliu: Since this is a Javadoc comment, would changing `name` to be `comment.block.javadoc.java` suffice?  Do you think there would still be a benefit to having a separate contentName?

grammars/java.cson
```diff
@@ -462,6 +462,46 @@
   'comments-inline':
     'patterns': [
       {
+        'begin': '^\\s*/\\*\\*'
+        'captures':
```
50Wliu: Please make this a separate `beginCaptures` and `endCaptures`.

grammars/java.cson
```diff
@@ -462,6 +462,46 @@
...
+        'end': '\\*/'
+        'name': 'comment.block.java'
+        'patterns': [
+          {
+            'captures':
+              '1':
+                'name': 'keyword.other.documentation.javadoc.java'
+              '2':
+                'name': 'variable.parameter.java'
+            'match': '(@param)\\s+(\\S+)'
```
50Wliu: `match` should go above `captures`

50Wliu:
> I tried adding highlighting for embedded code snippet in javadoc comments, but could not make it work.

Can you give an example of an embedded code snippet?  I might have some ideas for how to highlight that.

grammars/java.cson
noseglid: Changing to `comment.block.javadoc.java` would work well I think

sadikovi:
Thank you @noseglid @50Wliu, I will address comments later today. It seems like my first approach was suboptimal, I should have created a separate pattern for it.

sadikovi:
@50Wliu It is usually done using `{@code ... }`, can be multi-line, I do not really use this feature, I am not sure how often people use it, but it is like this:
```java
/**
 * Usage:
 * {@code
 * int s = getSomething();
 * boolean a = s > 1;
 * }
 */
public static int getSomething() {
  return -1;
}
```

When I was trying it, I could not make it to treat `*` on every line as a beginning of a comment.
Anyway, I will refactor code as @noseglid suggested, this should make it easier to try fixing this.

50Wliu:
Hmm yeah, that'll be tricky.  When I have time I'll pull down this branch and see if I can figure out something that works.

sadikovi:
@noseglid @50Wliu I addressed your comments, added `comments-javadoc` pattern that is applied inside class-body and at root level only, there is a test for comment inside the method.

I am a little bit confused with:
```
'name': 'comment.block.javadoc.java'
'contentName': 'meta.comment.block.javadoc.java'
```

Should I keep it this way, or should I change it to:
```
'name': 'meta.comment.javadoc.java'
'contentName': 'meta.comment.javadoc.body.java'
```
similar to **class** and **method**, or any other way?

Have not tried multi-line `{@code ...}`, and honestly, I do not think it is very essential to add that.

noseglid:
I think you should remove `contentName` leaving only `name: 'comment.block.javadoc.java'`.

Other than that this looks nice, I think.

sadikovi:
Ok, will do, thanks.

sadikovi:
@50Wliu @noseglid Rebased PR. Could you review it again, please?

noseglid:
LGTM 👍

sadikovi:
@50Wliu can you have a look at the PR?

50Wliu:
Been busy but I'll probably have some free time over the weekend or next week.

grammars/java.cson
```diff
@@ -495,6 +501,50 @@
...
+        {
+          'match': '(@(?:exception|throws))\\s+(\\S+)'
+          'captures':
+            '1':
+              'name': 'keyword.other.documentation.javadoc.java'
+            '2':
+              'name': 'entity.name.type.class.java'
+        }
+        {
+          'match': '\\{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]*\\)).*\\}'
```
50Wliu: Very small nit: the `{` and `}` don't need to be escaped.

spec/java-spec.coffee
```diff
@@ -633,3 +633,172 @@ describe 'Java grammar', ->
     expect(lines[8][10]).toEqual value: ' ', scopes: scopeStack
     expect(lines[8][11]).toEqual value: '{', scopes: scopeStack.concat ['punctuation.section.catch.begin.bracket.curly.java']
     expect(lines[8][12]).toEqual value: '}', scopes: scopeStack.concat ['punctuation.section.catch.end.bracket.curly.java']
+
+  it 'tokenizes comment inside method body', ->
+    lines = grammar.tokenizeLines '''
+      class Test
+      {
+        private void method() {
+          /** sort of javadoc comment */
```
50Wliu: For clarification can you rename this to `invalid javadoc comment`?

sadikovi:
@50Wliu I addressed your comments, would you mind taking a look again?

50Wliu:
Thanks!

sadikovi:
@50Wliu @noseglid Thank you very much for reviewing and merging pull request!

PR #58, opened 2016-06-13
Variables structure take two

noseglid:
Main problem is the identification of variables.
It was too lenient in the first take because it didn't require the a space to exist (e.g. between `int` and `name` in `int name;`).

I am not super happy with the current solution (repeating the "generics, dots, commas blalba" regex and check for a space in between), but I'm at a loss for better alternatives. The way it was before didn't work with generics (see comment in #47). At least this works in all cases.

The issue in #47 is different with this PR. E.g. example (row 13) below gets
  * `meta.definition.variable.java storage.type.java` on `Rectangle2D` (correct) 
  * `meta.definition.variable.java punctuation.separator.period.java` on `.` (correct)
  * `meta.definition.variable.java variable.other.property.java` on `Double` (while correct, I'd like to se `storage.type.java` here too)
  * `meta.definition.variable.java variable.other.definition.java` on `d` (correct)

So much better, just a minor wrinkle to iron out.

Example:
screen shot 2016-06-13 at 2 23 17 pm

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1118,10 +1118,11 @@
         (
           (void|boolean|byte|char|short|int|float|long|double)
           |
-          (([a-z]\\w*\\s*\\.\\s*)*[A-Z]+\\w*) # e.g. javax.ws.rs.Response
+          ((\\w+\\.)*[A-Z]+\\w*) # e.g. `javax.ws.rs.Response`, or `String`
         )
-        \\s*
-        [\\w\\d_<>\\[\\],\\?][\\w\\d_<>\\[\\],\\?\\s]*
+        ([\\w\\d<>\\[\\],?][\\w\\d_<>\\[\\],?\\s]*)?
+        \\s+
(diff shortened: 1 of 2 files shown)


50Wliu:
Tested this out and couldn't find anything wrong!  Thanks once again @noseglid!

50Wliu:
And of course the moment I hit the merge button I see something:
```java
something.add(0, new Integer(((variable >= 0) ? 1 : -1)));
```
Anything inside `Integer()` isn't highlighted.

50Wliu:
One more:
```java
else if(testDistanceX < 0 && testDistanceX <= testDistanceY) // Too close to the east/west wall
```
Everything starting from the `D` in the second `testDistanceX` is being tokenized as `meta.definition.variable.java`.

I really need to test out these PRs more before merging them :/.

noseglid:
I didn't catch that either, and I spent a whole day working with it enabled. I'll look into it.

PR #59, opened 2016-06-14
Require storage type and identifier

noseglid:
To avoid matching strings like `Something <= OtherThing` where the
`<` of the comparison operator was mistaken for a generic, require
there to be at least one storage type and at least one identifier.

Dollar are valid identifiers, added that in a few places it was
missing.

Also simplifies some regexes. E.g. `\w` matches `a-zA-Z0-9_` so
some redundant chars could be removed.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1123,7 +1123,7 @@
         ([\\w<>\\[\\],][\\w<>\\[\\],?\\s]*)?
         \\s+
         [A-Za-z_$][\\w$]* # At least one identifier after space
-        ([\\w\\[\\],$][\\w\\[\\],\\s]*)? # possibly more generic, primite arrays etc.
+        ([\\w\\[\\],$][\\w\\[\\],\\s]*)? # possibly primitive array or additional identifiers
         \\s*(=|;)
       ))
     '''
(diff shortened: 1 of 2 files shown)


noseglid:
Sample:
screen shot 2016-06-14 at 3 58 41 pm

grammars/java.cson
```diff
@@ -1120,17 +1120,18 @@
           |
           ((\\w+\\.)*[A-Z]+\\w*) # e.g. `javax.ws.rs.Response`, or `String`
         )
-        ([\\w\\d<>\\[\\],?][\\w\\d_<>\\[\\],?\\s]*)?
+        ([\\w\\d<>\\[\\],][\\w\\d<>\\[\\],?\\s]*)?
```
MaximSokolov: `\d` is also redundant as `\w` includes digits. Same for below

noseglid: Right you are!

50Wliu:
@noseglid Can you add some regression tests for this and #58?  I noticed that for the previous two PRs none of the specs caught anything wrong.

noseglid:
Sure. I've added some specs for most of the cases.

50Wliu:
Ran this through another quick eyeball using the same file that I've been seeing all the issues on and couldn't find anything wrong.  Here's hoping I didn't miss anything...

noseglid:
Rather quick to deploy and quick to fix than slow to both :)

PR #61, opened 2016-06-22
Fix `final` as modifier in method parameters

sadikovi:
This PR fixes issue when method parameter name that starts with `final`, e.g. `finalScore` would be highlighted as `storage.modifier.java` alongside with actual modifier in front of type. Note that this only occurs inside class body.

Here is an example:
sample-1

Patch fixes it by checking word boundary for `final` modifier keyword. In cases where it is part of the variable name, the correct `variable.parameter.java` is assigned to it.
After fix:
sample-2

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -990,7 +990,7 @@
     'patterns': [
       {
         # `final` keyword must have at least one subsequent space before variable type
-        'match': '(final)\\s+'
+        'match': '\\b(final)\\b'
         'captures':
           '1':
             'name': 'storage.modifier.java'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -989,10 +989,8 @@
   'parameters':
     'patterns': [
       {
-        'match': '\\b(final)\\b'
-        'captures':
-          '1':
-            'name': 'storage.modifier.java'
+        'match': '\\bfinal\\b'
+        'name': 'storage.modifier.java'
       }
       {
         'include': '#annotations'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -989,7 +989,6 @@
   'parameters':
     'patterns': [
       {
-        # `final` keyword must have at least one subsequent space before variable type
         'match': '\\b(final)\\b'
         'captures':
           '1':
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -989,8 +989,11 @@
   'parameters':
     'patterns': [
       {
-        'match': 'final'
-        'name': 'storage.modifier.java'
+        # `final` keyword must have at least one subsequent space before variable type
+        'match': '(final)\\s+'
+        'captures':
+          '1':
+            'name': 'storage.modifier.java'
       }
       {
         'include': '#annotations'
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu @noseglid Could you, please, have a look at PR?
I just noticed this highlighting today when debugging some of my code.

grammars/java.cson
```diff
@@ -989,8 +989,11 @@
   'parameters':
     'patterns': [
       {
-        'match': 'final'
-        'name': 'storage.modifier.java'
+        # `final` keyword must have at least one subsequent space before variable type
+        'match': '(final)\\s+'
```
noseglid: Alternatively, wich is used more throughout the file is to use the word boundary `\b`. I'm not sure if this fixes a usage like `final int somefinal ` (where there are spaces after `final` but not before. Word boundary would definitiely solve that too.

sadikovi: I will try and update according to your suggestion. I also started thinking that my fix is not great, I think we should enforce 'final' to be the first word preceding type and variable name, I will play with it today.

noseglid: hmm, i think both `final private int` and `private final int` are allowed. Keep that in mind.

sadikovi: Yeah, you are right:)

50Wliu:
`\\bfinal\\b` should work.

sadikovi:
@50Wliu @noseglid I updated pattern for `final`, also added test cases that @noseglid suggested. Could you review again? Thanks.

grammars/java.cson
```diff
@@ -989,8 +989,10 @@
   'parameters':
     'patterns': [
       {
-        'match': 'final'
-        'name': 'storage.modifier.java'
+        'match': '\\b(final)\\b'
+        'captures':
+          '1':
+            'name': 'storage.modifier.java'
```
noseglid: There's no need for parenthesis or `captures` anymore. Just use `name`.

sadikovi: Yep, will do, thanks.

noseglid:
This looks good to me 👍

spec/java-spec.coffee
```diff
@@ -235,6 +235,41 @@ describe 'Java grammar', ->
     expect(lines[3][1]).toEqual value: '{', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.begin.bracket.curly.java']
     expect(lines[4][1]).toEqual value: '}', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'punctuation.section.method.end.bracket.curly.java']
 
+  it 'tokenizes `final` in inline method parameter', ->
+    {tokens} = grammar.tokenizeLine 'public int doSomething(final int finalScore) { return finalScore; }'
```
50Wliu: This test appears to be basically the same thing as the one in line 246?

sadikovi: What I found was that highlighting worked correctly in case of a test on line 238 and did not work in case of test on line 245, so I added both. Do you think that L238 test should be removed?

50Wliu: Well the thing is this is technically invalid Java, since you can't have a function definition without a class definition as well.  If you mean to test a function call instead (which is what it's currently doing), I would do that, or if the name of the test is still what you want, then you would need to wrap a class around it.

sadikovi: Yes, you are right, I was just experimenting with different scenarios where it worked and did not work. I will remove test, as it is invalid and essentially is a duplicate of test below.

sadikovi:
@50Wliu updated according to your comment - removed `inline` test and updated `class method` test.

PR #63, opened 2016-08-21
Fix inline comment in method signature

### LINKED ISSUE #60: Doesn't gray out comment blocks inside method/constructor parameter list
This snippet works correctly (comment is grayed out):
```(/* comment */)```

This one does not (comment is not grayed out):
```method(/* comment */)```

Full example with bug:
```public Constructor(String p1, String p2/*, String p3*/)```

Full example without bug:
```public (String p1, String p2/*, String p3*/)```

### PR DESCRIPTION
**Pull request by sadikovi**
This PR fixes bug with displaying inline comment that is mixed with method parameters + when there is just comment without method parameters, test cases are shown below:

inline-comment-method

Change is just adding `comments-inline` to the list of patterns to check in `methods`.

Fixes #60.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -884,7 +884,7 @@
         'patterns': [
           {
             'include': '#parameters'
-          },
+          }
           {
             'include': '#comments-inline'
           }
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -884,6 +884,9 @@
         'patterns': [
           {
             'include': '#parameters'
+          },
+          {
+            'include': '#comments-inline'
           }
         ]
       }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sadikovi** [comment]
@50Wliu @noseglid please review.

**Comment by noseglid** [comment]
I'd add checks for the surrounding tokens to (`public`, `void`, etc). At least on one of the tests.
Otherwise it LGTM

**Comment by sadikovi** [comment]
@noseglid thanks. Will do.

**Comment by sadikovi** [comment]
@noseglid addressed your comment.

**Comment by noseglid** [comment]
👍 LGTM

File: grammars/java.cson [hunk 1]
```diff
@@ -884,6 +884,9 @@
         'patterns': [
           {
             'include': '#parameters'
+          },
```
50Wliu: No trailing comma

sadikovi: Thanks, will fix.

sadikovi:
@50Wliu fixed your comment, could you review again?

50Wliu:
Thanks!

sadikovi:
@50Wliu @noseglid thank you for reviewing and merging PR!

Issue #64, opened 2016-09-06
Add patterns for JSP expression language (EL)


MoritzKn:
Currently the JSP expression language just includes the normal java syntax ([See jsp grammars line 80](https://github.com/atom/language-java/blob/master/grammars/java%20server%20pages%20%28jsp%29.cson#L80)), but actually JSP EL has a quit different syntax then Java.

See: [JSP EL on tutorialspoint](http://www.tutorialspoint.com/jsp/jsp_expression_language.htm) or the [spec](http://download.oracle.com/otn-pub/jcp/jsp-2.1-fr-spec-oth-JSpec/jsp-2_1-fr-spec-el.pdf) (probably _a bit_ overkill)

A few exampel JSP EL snippets :

``` jspx
${foo eq bar}
```

``` jspx
${fn:contains(true, 'x') ? bar : 'foo'}
```

``` jspx
${bar mod 10 eq 0}
```

``` jspx
${bar('some longer string', bar())}
```

``` jspx
${param["username"]}
```

``` jspx
${fn:indexOf(pageContext.session.getAttribute('bar'), 'foo') eq -1}
```

MoritzKn:
I think about doing this myself but I may need a little bit help.

MoritzKn:
With #65 is is resolved now

PR #65, opened 2016-09-08
Add Unified Expression Language (EL) and include it in JSP

MoritzKn:
This is my attempt to implement #64. 
Because other technologies (like JSF) use EL as well, the EL grammar is added as its own language file.

There are a few things I am unsure about:
- Is `source.java.el` a good scopeName?  
  The problem I see is that snippets, autocomplete providers and alike which are intended to work with Java will also apply to EL.
- Should I add test and if so, how?

Some JSP code I used for testing: 

``` jsp
${foo}
${fooBar}
${42}
${foo42}
${42foo}
${fooeq}
${foodiv}
${foomod}
${notfoo}
${footrue}
${42+4}
${42==4}
${'bar \' \\ sdaff " dsaf'}
${"dsa f ' 'dsaf sdf \\ \" asd fds af"}
${"foo"eq"foo"}
${foo eq bar}
${foo == bar}
${foo==bar}
${foo ne bar}
${foo !=bar}
${0 lt 5}
${0 < 5}
${8 gt 5}
${8 > 5}
${80 le 5}
${80 <= 5}
${42 ge 0}
${42 >= 0}
${42 + 0}
${42 - 0}
${42 * 0}
${42 div 0}
${42 / 0}
${42 mod 0}
${42 % 0}
${not bar}
${!bar}
${"foo" eq 'foo'}
${"\"'" eq '"\''}
${true}
${false}
${fn:contains('foo', 'bar')}
${foo ? 'foo' : 'bar'}
${foo['bar']}
${foo[42]}
${(42 lt 5) and (45 ne 8)}
${fn:indexOf('fooo', 'o')}
${not empty bar and fn:contains(bar, 'foo')}

<c:if test="${not empty bar and fn:contains(bar, 'foo')}">
    ${bar}
    ${setFoo(true)}
</c:if>
```

### DIFF
--- a/grammars/unified expression language (el).cson
+++ b/grammars/unified expression language (el).cson
@@ -1,95 +1,88 @@
 'scopeName': 'source.java.el'
 'patterns': [
   {
-    'include': '#el'
+    'name': 'keyword.control.ternary.java.el'
+    'match': '\\?|(?<=\\s):'
   }
-]
-'repository':
-  'el':
+  {
+    'name': 'keyword.operator.comparison.java.el'
+    'match': '((==|!=|<=|>=|<|>)|\\b(eq|ne|le|ge|lt|gt)\\b)'
+  }
+  {
+    'name': 'keyword.operator.empty.java.el'
+    'match': '\\b(empty)\\b'
+  }
+  {
+    'name': 'keyword.operator.arithmetic.java.el'
+    'match': '(?:(\\-|\\+|\\*|\\/|%)|\\b(div|mod)\\b)'
+  }
+  {
+    'name': 'keyword.operator.logical.java.el'
+    'match': '(?:(!|&&|\\|\\|)|\\b(not|and|or)\\b)'
+  }
+  {
+    'name': 'namespace.java.el'
+    'match': '[a-zA-Z]+(:)'
+    'captures':
+      '1':
+        'name': 'punctuation.separator.namespace.java.el'
+  }
+  {
+    'match': ','
+    'name': 'meta.delimiter.java.el'
+  }
+  {
+    'match': '\\(|\\)'
+    'name': 'meta.brace.round.java.el'
+  }
+  {
+    'match': '\\[|\\]'
+    'name': 'meta.brace.square.java.el'
+  }
+  {
+    'name': 'constant.boolean.java.el'
+    'match': '\\b(true|false)\\b'
+  }
+  {
+    'name': 'constant.null.java.el'
+    'match': '\\bnull\\b'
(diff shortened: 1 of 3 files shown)


MoritzKn:
Here a few EL examples from the oracle page: http://docs.oracle.com/javaee/6/tutorial/doc/bnaim.html#bnain

grammars/unified expression language (el).cson
```diff
@@ -0,0 +1,94 @@
+'scopeName': 'source.java.el'
+'name': 'Unified Expression Language (EL)'
```
50Wliu: Is this a grammar that someone will ever want to explicitly choose in the Grammar Selector?  If not, this line should be removed.

grammars/unified expression language (el).cson
```diff
@@ -0,0 +1,94 @@
...
+'fileTypes': []
```
50Wliu: :fire:

grammars/unified expression language (el).cson
```diff
@@ -0,0 +1,94 @@
...
+'patterns': [
+  {
+    'include': '#el'
```
50Wliu: Why not just include all the patterns here directly?

50Wliu:
Also yes, specs will be required for this PR.  I think some basic tests for each pattern should suffice.

grammars/unified expression language (el).cson
MoritzKn: > The primary new feature of JSP 2.1 is the unified expression language (unified EL), which represents a union of the expression language offered by JSP 2.0 and the expression language created for JavaServer Faces technology

-- https://docs.oracle.com/javaee/5/tutorial/doc/bnahq.html

Meaning the unified el grammar could also be useful outside of JSP in a JavaServer Faces (JSF) context. Therefore I thought a separate grammar would make sense although no JSF language package exists yet.

MoritzKn:
@50Wliu rebases against master, fixed your remarks and added specs.

grammars/unified expression language (el).cson
50Wliu: So does `#el` only contain the expression language up to JSP 2.0?  And that perhaps in the future someone may add support for JSP 2.1 in this file?

MoritzKn: Ups, I mistaken the code snippet for something else and comply misunderstood you. 🙈 
Ignore my first comment. You are right, it makes no sense to include everything like this. I probably just copied this from some other grammar without understanding what exactly it meant.

PR #66, opened 2016-09-27
Add JSP snippets

MoritzKn:
See: [JSP Syntax on Tutorialspoint](https://www.tutorialspoint.com/jsp/jsp_syntax.htm) and [JSP - Standard Tag Library on Tutorialspoint](https://www.tutorialspoint.com/jsp/jsp_standard_tag_library.htm)

### DIFF
--- a/snippets/language-java.cson
+++ b/snippets/language-java.cson
@@ -130,6 +130,30 @@
     'prefix': 'wh'
     'body': 'while ($1) {\n\t$0\n}'
 '.text.html.jsp':
+  'scriptlet':
+    'prefix': '%'
+    'body': '<%\n\t$1\n%>'
+  'scriptlet (xml)':
+    'prefix': 'scriptlet'
+    'body': '<jsp:scriptlet>\n\t$1\n</jsp:scriptlet>'
+  'declaration':
+    'prefix': '%!'
+    'body': '<%!\n\t$1\n%>'
+  'declaration (xml)':
+    'prefix': 'declaration'
+    'body': '<jsp:declaration>\n\t$1\n</jsp:declaration>'
+  'expression':
+    'prefix': '%='
+    'body': '<%=\n\t$1\n%>'
+  'expression (xml)':
+    'prefix': 'expression'
+    'body': '<jsp:expression>\n\t$1\n</jsp:expression>'
+  'comment':
+    'prefix': '%--'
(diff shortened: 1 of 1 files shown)


MoritzKn:
After using this for myself for some time I discovered a usability problem with the JSTL Tag Snippets:
For example
```jsp
<c:if<tab>
```
results in
```
<c:<c:if test="${true}">

</c:if>
```
because `if` is a snippet shortcut.
I don't know what would be a good alternative.. . Maybe `<c:if` instead of `if`, but that's pretty verbose.

MoritzKn:
Or is it possible to use a regex like `(?:^|[^<])if` as short cut?

MoritzKn:
I figured it would be nice if "tag snippets" would only work outside of tags. So that this:
```xml
<div out[tab]>
</div>
```
wouldn't result in this:
```xml
<div <c:out value="${}"/>>
</div>
```

50Wliu:
Hmm.  `c:if` should definitely take priority over just `if`.  I wonder where that matching occurs.

MoritzKn:
@50Wliu I've done some research on that `<c:if` vs `if` problem: 
The snippets package does -- probably for performance reasons -- some complicated stuff, but basically it just checks if the [text behind the cursor][0] is [equal][1] to the prefix of a [snippet in scope][3] and if that's the case, it [replaces the prefix with the snippet text][4].

Now I could do two things:
- Add a second version of each "tag snippets" whose prefix starts with `<c:`
- Add a second version of each "tag snippets" with out any prefix. That's what the [language-html package has done][5]. This will remove the snippets inside tags.

I prefer the first method, but I don't like that this means I have to duplicate all the code.

[0]: https://github.com/atom/snippets/blob/master/lib/snippets.coffee#L252
[1]: https://github.com/atom/snippets/blob/master/lib/snippets.coffee#L278
[3]: https://github.com/atom/snippets/blob/master/lib/snippets.coffee#L214
[4]: https://github.com/atom/snippets/blob/master/lib/snippet-expansion.coffee#L16
[5]: https://github.com/atom/language-html/blob/master/snippets/language-html.cson#L384

MoritzKn:
I fixed this problem in the last commit. It's not perfect because it adds some redundancy, but I think it's the best we can get. Correct me if I'm wrong.

PR #70, opened 2016-12-01
Overhaul numeric matching based on the offical specs

### LINKED ISSUE #69: syntax coloring is broken for some floating point constants
Hi,

java syntax coloring breaks for some floating point constants:

```java
float foo = 0.1f; // okay
float bar = .1f; // not okay
```

The latter case leads to '1f' being highlighted in bright red.

Pic attached to illustrate the issue.

atom-syntax

50Wliu:
http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.1
http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.2

Fixes #69

## TODO:
* [ ] Individual decimal scopes (maybe)
* [x] Specs
* [x] Thorough testing

/cc @MaximSokolov: You've helped rewrite some number highlighting before.  Mind taking a look at this as well?


50Wliu:
Ok.  I think I'm going to merge this without decimal scopes for now.

PR #74, opened 2016-12-27
Fix static initializer

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue with declaring static initializer as opposed to instance/method initializer. Currently if you write this in Java class:
```java
class Test {
  static HashSet<Integer> set = ...
  static {
    set.add(1);
  }
}
``` 
The inside of initializer is not highlighted correctly (see image below), and treated as partially written method syntax.

Note that instance/method initializers work correctly, pattern just does not understand `static` modifier prefix. I fixed by adding new pattern `static-initializer` that is built from `anonymous-block-and-instance-initializer` and `static` keyword, and added it into `class-body` scope. Also added tests for static/instance/method initializers.

before
after

### Alternate Designs
I considered first to add `static-initializer` totally separately with its own syntax, but that introduces more code changes and duplicates the effort of having two code blocks that handle the same functionality, really. I was also thinking about just modifying `anonymous-block-and-instance-initializer` itself to handle `static` in front of it, but thought that capturing optional group was not the best approach (could not see examples in existing code either).

### Benefits
Fixes highlighting of static initializers.

### Possible Drawbacks
Possibly breaks other parts of code, though it passes unit-tests.

### Applicable Issues
Currently there is a bug/feature of highlighting `static` inside method body. 
```java
class Test {
  public void func() {
    static {
      int a = 1;
    }
  }
}
```
This happens in master branch, compiler complains about this code. I am not sure if this is expected, and/or I need to fix it in this PR.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -299,6 +299,9 @@
         'include': '#generics'
       }
       {
+        'include': '#static-initializer'
+      }
+      {
         'include': '#methods'
       }
       {
@@ -326,6 +329,16 @@
     'patterns': [
       {
         'include': '#code'
+      }
+    ]
+  'static-initializer':
+    'patterns': [
+      {
+        'include': '#anonymous-block-and-instance-initializer'
+      }
+      {
+        'match': 'static'
+        'name': 'storage.modifier.java'
       }
     ]
   'code':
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu @noseglid Could you please review this PR? Thanks!

sadikovi:
Hmm... appveyor `ATOM_CHANNEL=stable` failed with this:
```
Downloading latest Atom release...
Error downloading remote file: One or more errors occurred.
Inner Exception: Remote server returned 503: first byte timeout
Command exited with code 
```
`ATOM_CHANNEL=beta` succeeded, error above does not seem to be related to this PR.

50Wliu:
Looks good, and thanks for the extra anonymous block tests!
I'm happy with this approach, especially since it's not possible to handle the case where the opening brace is on a new line if you tried using an optional capture in `anonymous-block-and-instance-initializer`.

sadikovi:
@50Wliu thank you for merging this!

PR #75, opened 2017-01-31
Attempt to fix overlap of generics and if-else block

sadikovi:
### Requirements

### Description of the Change
In released version of Atom 1.13.0 and in current master branch there is an issue when expression inside `if (...)` is parsed as generics when capitalized variable is used with `<` and some other combinations, see screen shots below. This can even affect all subsequent highlighting as shown on image below (with string). For most part it is merely minor, out of ordinary highlighting of variables, not critical at all.

java-generic-overlap
java-generic-overlap-3

- Examples that break hightlighting: `if (A < a) { ... }`, `if (A < a && b < a) { ... }`, and `if (A < a || b > a) { ... }`
- Examples that work: `if (a < A) { ... }`, `if (a < b) { ... }`

I tried fixing most of the cases, currently I make sure that generic ends up on the same line. From what I have seen people keep generics expression on single line, and thought it would be a fair assumption to make. Code **does not fix** situations with logical operators, e.g. `&&`, `||`.  I also added tests for generics, which I think could be useful even if PR is closed (I can add tests separately).

After:
java-generic-overlap-after-1
java-generic-overlap-after-2

### Alternate Designs
As an alternative approach I was considering capturing `condition-block` similar to other patterns like `try-catch`, so we know that everything inside that block can be variable names and/or operators only. This would fix all cases above. I did not make the change for a reason of not being such critical at the moment and also wanted to hear from community, and discuss/approve approaches.  

### Benefits
Current solution is simple, alternative would require more code, but it would cover all cases and potentially more robust to further changes (+ easy to maintain over long regex).

### Possible Drawbacks
My change might affect some other untested cases.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -912,7 +912,8 @@
       }
       {
         # Match possible generics first, eg Wow<String> instead of just Wow
-        'begin': '\\b(?:[a-z]\\w*\\s*(\\.)\\s*)*[A-Z]+\\w*\\b\\s*(?=<)'
+        # Also expression ensures that generic is specified on single line with closing '>'
+        'begin': '\\b(?:[a-z]\\w*\\s*(\\.)\\s*)*[A-Z]+\\w*\\b\\s*(?=<.*>)'
         'beginCaptures':
           '0':
             'name': 'storage.type.java'
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -451,6 +451,79 @@ describe 'Java grammar', ->
     expect(lines[6][3]).toEqual value: 'Object', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.return-type.java', 'storage.type.java']
     expect(lines[6][5]).toEqual value: 'otherMethod', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.identifier.java', 'entity.name.function.java']
 
+  it 'tokenizes generics in if-else code block', ->
+    lines = grammar.tokenizeLines '''
+      void func() {
+        int a = 1, A = 2, b = 0;
+        if (A < a) {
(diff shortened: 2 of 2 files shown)


sadikovi:
@50Wliu @noseglid could you review please? Also would really appreciate your opinion on my approach + alternative or other options, e.g. how to fix cases with `&&` and `||` I mentioned above.

50Wliu:
I always prefer a more robust solution when one is available.  So I would be in favor of creating a new `control-block` group (or similar) that does not include generics.

sadikovi:
Thanks @50Wliu, will try adding `control-block` group.

grammars/java.cson
```diff
@@ -912,7 +912,8 @@
       }
       {
         # Match possible generics first, eg Wow<String> instead of just Wow
-        'begin': '\\b(?:[a-z]\\w*\\s*(\\.)\\s*)*[A-Z]+\\w*\\b\\s*(?=<)'
+        # Also expression ensures that generic is specified on single line with closing '>'
+        'begin': '\\b(?:[a-z]\\w*\\s*(\\.)\\s*)*[A-Z]+\\w*\\b\\s*(?=<.*>)'
```
sadikovi: It is actually even the wrong fix - you can have generic expression split into multiple lines.

sadikovi:
It seems like it does not just affect `if-else` statements, it is pretty much anything that can trigger generics, e.g. 
```
boolean flag;
flag = A < a && str.equals("str") && a > A;
```

50Wliu:
Hmm.  Let me see if I can tweak the generics code then.

50Wliu:
Hmm yes, this will be very tricky.  I'm not sure it'll be possible, actually.

sadikovi:
I am going to close this for now, meanwhile will investigate. My thoughts were updating generics pattern to not include && or || and " somehow.

Thanks for the feedback @50Wliu!

PR #78, opened 2017-02-03
Check to make sure an identifier exists before tokenizing something as a storage type

### LINKED ISSUE #76: Capital variables tokenized as storage types
### Prerequisites

### Description

When using capital variables in code, they are tokenized incorrectly as storage types.  Currently, they shouldn't be tokenized as anything.
storage-type-variables

### Steps to Reproduce

1. Open a Java file and create the class/method boilerplate
2. `int G = 0;` inside a method
3. `G += 3;`

**Expected behavior:** G on the third line shouldn't be tokenized as anything special

**Actual behavior:** G is tokenized as `storage.type.java`

**Reproduces how often:** 100%

### Versions

Atom version doesn't matter.  I'm running language-java commit d9b52c2.

50Wliu:
### Requirements

### Description of the Change

Before, any variable that started with an uppercase letter (eg `Hello`) would be considered a storage type, even if there was no identifier following it (eg normal variable usage, `Hello = 3`).  This PR makes sure that an identifier follows a potential storage type before tokenizing it as one.

### Alternate Designs

None.

### Benefits

People who use uppercase variables will be happy.

### Possible Drawbacks

Multiline variable declarations with an object type no longer have their storage type tokenized.  I think this is a fair compromise, considering that the variable itself already isn't tokenized.
```java
class Test {
  String
  stuff = 0; // String and stuff won't be highlighted
```

### Applicable Issues

Fixes #76, depends on #77

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -949,7 +949,7 @@
         'include': '#generics'
       }
       {
-        'begin': '\\b(?:[a-z]\\w*\\s*(\\.)\\s*)*([A-Z]+\\w*)\\b\\s*(?=\\[)'
+        'begin': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*([A-Z]\\w*)\\s*(?=\\[)'
         'beginCaptures':
           '1':
             'name': 'punctuation.separator.period.java'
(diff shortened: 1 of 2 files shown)


50Wliu:
I don't believe this is the correct fix.

PR #93, opened 2017-05-17
Attempt to fix overlap of generics 2

sadikovi:
### Requirements

### Description of the Change
In released version of Atom 1.17.0 and in current master branch there is an issue when expression inside `if (...)` is parsed as generics when capitalized variable is used with `<` and some other combinations, see screen shots below. This is also observed when bitwise operators are used. Result is that such _partial generic_ affects all subsequent highlighting (including other functions declarations, etc) as shown on image below. For most part it is merely minor rather annoying, out of ordinary highlighting of variables, not critical at all.

- Examples that break highlighting: `if (A < a) { ... }`, `if (A < a && b < a) { ... }`, and `if (A < a || b > a) { ... }`, and `t = M << x`
- Examples that work: `if (a < A) { ... }`, `if (a < b) { ... }`

I tried fixing most of the cases, currently I make sure that pattern for object types terminates when `;` is found, which is basically replacement of `(?<=>)` with `(?<!;)`. I am still not sure why it works, and would appreciate if someone could explain. From manual tests, it looks like this fixes situations with usage `&&` and `||` in `if-else` statements, which my previous PR reported as not covered.

### Before
before1

before2

### After
after1

after2

### Alternate Designs
Some of the designs were proposed in #75, but this solution looks fairly simple and does not break existing tests.

### Benefits
Current solution is simple, alternative would require more code, with potentially refactoring large portion of syntax highlighting. This solution also preserves all patterns for generics, and does not have constraint of my previous pull request (generic on single line only).

### Possible Drawbacks
The change might break some other untested cases.


sadikovi:
@50Wliu @noseglid could you review this pull request? This is the second take on generics after #75.

grammars/java.cson
```diff
@@ -962,13 +962,14 @@
       }
       {
         # Match possible generics first, eg Wow<String> instead of just Wow
-        'begin': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\s*(?=<)'
+        'begin': '(\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*)\\s*(?=<)'
```
sadikovi: This is done, so parsed `storage.type.java` does not include trailing spaces.

sadikovi:
@50Wliu could you review this pull request, please? Thanks!

50Wliu:
I must say that I'm still unhappy with this solution, however it may be the best option available.  It's been a while since I last looked at this, so give me a few days to try to get reassociated with the code.

sadikovi:
When debugging this problem, I noted that genetics pattern was correct, but generic termination capturing was somehow loose, therefore I decided to essentially enforce rule of terminating generic after `;`. Added tests to make sure new cases were covered, and old test cases passed.

I am happy to try a different approach to this problem, if you want.

sadikovi:
@50Wliu did you have a chance to look at this problem and pull request? I am not sure what the next step should be. Should I close this PR and try coming up with new solution or should we just mark it as known problem? Thanks.

sadikovi:
@50Wliu I also found that this code:
```java
class A<T> {
  private B<T>[] arr;
}
```

is not highlighted correctly in master branch (latest commit 652e8ba41a43d21fc92477da699fb99a655c7908), which I might have broken.

Found that this PR fixes this problem and returns correct highlight:)

Should I add test for this issue?

50Wliu:
Yes please, if it isn't already tested!  Sorry for the late reply.

sadikovi:
@50Wliu I was wrong, this PR does not fix the problem of tokenising generic arrays. I will create a new pull request to patch that issue.

sadikovi:
I created PR to fix aforementioned issue: https://github.com/atom/language-java/pull/110

sadikovi:
@50Wliu I rebased to master. Passes all tests locally. Should we proceed with this fix or should we address this issue differently?

50Wliu:
I want to apologize for the continued delays.  I really want to review this, I just haven't had enough time to do so.  Sorry to keep you waiting.

sadikovi:
@50Wliu It is okay, no worries, I am just curious what we do about the issue and resolution. It does not seem to be particularly common, though examples in description are still reproducible in the latest master branch.

grammars/java.cson
50Wliu: Move the opening parentheses to be after the `\\b`.

grammars/java.cson
```diff
@@ -962,13 +962,14 @@
...
         'beginCaptures':
-          '0':
-            'name': 'storage.type.java'
           '1':
+            'name': 'storage.type.java'
+          '2':
             'name': 'punctuation.separator.period.java'
-        'end': '(?<=>)'
+        # match until the line terminator, pattern (?<=>) breaks subsequent code highlighting
```
50Wliu: I think this comment needs to be updated because we're making sure there _isn't_ a `;` behind us.

sadikovi: Yes, you are right. I tested `(?!;)` and `(?<!;)`, and both pass tests. Should go with lookahead or lookbehind? Or does it mean that I should add more tests?

sadikovi:
@50Wliu, Thanks for the review! I rebased and addressed your comments. I left `(?<!;)` lookbehind, instead of using lookahead, but it looks like either works.

grammars/java.cson
50Wliu: I usually prefer lookaheads since they're easier to reason about.  Then I would also change `line terminator` to `statement terminator`.

sadikovi:
@50Wliu Thanks for patching this. Apologies - I somehow missed your answer in comments.

50Wliu:
One last thing - would it be more efficient to use `(?<=>)|(?!;)`?  That way we can exit the pattern earlier if a generic is detected.

sadikovi:
It passes tests, and I manually ran several examples - I agree, let's use `(?<=>)|(?!;)` instead of just `(?!;)`.

sadikovi:
@50Wliu Are you going to update the pattern? Otherwise, I could push commit, if you want.

50Wliu:
I'll update it :).

50Wliu:
Thanks for your patience with this 🙇

sadikovi:
Thanks for merging!

PR #110, opened 2017-09-17
Fix tokenising for generic arrays

sadikovi:
### Requirements

### Description of the Change
This pull request fixes issue with tokenising generics array. Example of code is below:
```java
class A<T> {
  private B<T>[] arr;
}
```

The reason why it was broken is variable pattern that now considers either generic declaration or primitive array (`[]` to be precise). The solution is combining those pattern into a chain rather than OR.

Before this patch:
before

After:
after

### Alternate Designs
Only one approach was considered - modifying `variables` pattern to handle such highlighting. Note that this is not very common in Java to use generics array like this, but is still worth fixing.

### Benefits
Fixes broken highlight. Works with my generics patch and does not break tests in master + in that patch.

### Possible Drawbacks
Might break other untested corner cases.

### Applicable Issues
None


sadikovi:
@50Wliu please review. Let me know if this solution could potentially break other cases.

sadikovi:
@50Wliu Could you have a look at this PR? Thanks!

50Wliu:
Whoops, I didn't get a notification for this PR for some reason!

sadikovi:
@50Wliu Thanks for merging!

Issue #112, opened 2017-10-14
Scope `@` in annotations as punctuation


chbk:
Can `@` have its own scope? It would be handy as `punctuation.definition.annotation`, similarly to `$` scoped as `punctuation.definition.variable` in php for example.

PR #113, opened 2017-10-18
Tokenize `@`

chbk:
- Tokenizes `@` as `punctuation.definition.annotation`, as per #112

- Prevents tokenization of `@interface` in class definitions. It is used to define an annotation, not a regular interface.

    In the examples below, `interface` is now `meta.declaration.annotation` >  `storage.modifier`:

    ```java
    @interface Author {
      String name();
    }

    public @interface Author {
      String name();
    }
    ```

    In the examples below, `interface` is still `meta.class` > `meta.class.identifier` > `storage.modifier`:

    ```java
    interface Author {
      String name();
    }

    public interface Author {
      String name();
    }
    ```

- Adds specs

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -122,7 +122,7 @@
   'annotations':
     'patterns': [
       {
-        'begin': '((@)[^ (]+)(\\()'
+        'begin': '((@)[^\\s(]+)(\\()'
         'beginCaptures':
           '1':
             'name': 'storage.type.annotation.java'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -122,11 +122,13 @@
   'annotations':
     'patterns': [
       {
-        'begin': '(@[^ (]+)(\\()'
+        'begin': '((@)[^ (]+)(\\()'
         'beginCaptures':
           '1':
             'name': 'storage.type.annotation.java'
           '2':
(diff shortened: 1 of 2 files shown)


grammars/java.cson
```diff
@@ -122,11 +122,13 @@
   'annotations':
     'patterns': [
       {
-        'begin': '(@[^ (]+)(\\()'
+        'begin': '((@)[^ (]+)(\\()'
```
50Wliu: I would rather we use `\\s` here instead of <code> </code>.

grammars/java.cson
```diff
@@ -122,11 +122,13 @@
...
         'beginCaptures':
           '1':
             'name': 'storage.type.annotation.java'
           '2':
+            'name': 'punctuation.definition.annotation.java'
+          '3':
             'name': 'punctuation.definition.annotation-arguments.begin.bracket.round.java'
         'end': '(\\))'
```
50Wliu: Nit: capture can be removed and the 0th capture can be targeted instead.

chbk: Can you clarify?

50Wliu:
```cson
'end': '\\)'
'endCaptures':
  '0':
    'name': # ...
```

PR #119, opened 2018-01-04
Added try-catch Snippet

danielct1995:
Just added a generic try-catch. It's declared with a prefix named 'trCa'

### DIFF
--- a/snippets/language-java.cson
+++ b/snippets/language-java.cson
@@ -73,7 +73,7 @@
     'body': 'try {\n\t$0\n}'
   'try-catch':
     'prefix':'trCa'
-    'body': 'try{$1}catch(${2:Exception} ${3:e}){$4}'
+    'body': 'try{\n\t$1\n}catch(${2:Exception} ${3:e}){\n\t$4\n}'
   'variable':
     'prefix': 'v'
     'body': '${1:String} ${2:var}${3: = ${0:null}};'
--- a/snippets/language-java.cson
+++ b/snippets/language-java.cson
@@ -72,8 +72,8 @@
     'prefix': 'tr'
     'body': 'try {\n\t$0\n}'
   'try-catch':
-    'prefix':'trCa'
-    'body': 'try{\n\t$1\n}catch(${2:Exception} ${3:e}){\n\t$4\n}'
+    'prefix':'trc'
+    'body': 'try {\n\t$1\n} catch(${2:Exception} ${3:e}) {\n\t$4\n}'
   'variable':
     'prefix': 'v'
     'body': '${1:String} ${2:var}${3: = ${0:null}};'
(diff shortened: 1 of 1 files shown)


snippets/language-java.cson
```diff
@@ -71,6 +71,9 @@
   'try':
     'prefix': 'tr'
     'body': 'try {\n\t$0\n}'
+  'try-catch':
+    'prefix':'trCa'
```
sadikovi: I was thinking if it is worth naming it `trca` or `tryc` - looks like all snippet prefixes are lowercase, and just easier to type:). @danielct1995 what do you think? By the way, I do not know if there is any convention around naming snippets:).

I would also refactor pattern so it looks something like this:
```
  'try-catch':
    'prefix':'trca'
    'body': 'try {\n\t$1\n} catch (${2:Exception} ${3:e}) {\n\t$4\n}'
```

which would render it like this:
```java
try {
..
} catch (Exception e) {
..
}
```

I think this is more or less how people would want it, though either way it is fine!

sadikovi:
Another idea: would it be useful to add try-catch-finally as snippet?

snippets/language-java.cson
danielct1995: Hello @sadikovi , first i want to thank your comments.
About the prefix I must say that in general I don't agree the way that are made most part of this.
In my opinion, they are extremely short and don't seize atom's auto-completion feature, which allows us to declare a little bit longer and explanatory prefix to each snippet. This kind of snippets should work like Eclipse.
Also the auto-completion feature ignores the uppercasses so both 'trca' and 'trCa' works.

On the other hand, I have to thank your input, it would be definitely better to render the snippet as the way you told.

sadikovi: Good point about auto completion, I actually have not tested that:).
Even though I agree that snippet names should be more self-explanatory, having short names allows to type less:).

I am happy with either name, but I was actually considering `trc` for try-catch and `trcf` for try-catch-finally.

Thanks for changing snippet body, though I am not sure if you want to add spaces before { and after }, like in my example, - this would make it similar to `try` snippet.

sadikovi:
Thanks for updating the code! 

@50Wliu could you have a looking at this PR? Thanks!

snippets/language-java.cson
danielct1995: Hello @sadikovi, if you are considering to create a try-catch-finally snippet it would be a great idea to change prefix to trc for 'Try-catch' snippet.

sadikovi: Never mind try catch finally. Could you change name of the prefix to `trc` or `trca` and add white spaces around try, catch key words? Thanks.

danielct1995: Of course, @sadikovi. It's done !

sadikovi:
LGTM. Thanks for fixing it!

sadikovi:
@danielct1995 thanks for your work! 

@50Wliu thanks for the review!

PR #121, opened 2018-01-04
Fix issue when tokenizing an empty comment

### LINKED ISSUE #120: Syntax highlighting treats `/**/` as the start of a JavaDoc

### Description

When using the Java language, if a multiline comment has no inner content, the language assumes the comment never closed.

### Steps to Reproduce

Paste the following code into atom.

```java
/**/  int nohilighting  = 1;
/**/  int yeshilighting = 1;
/**/  int nohilighting  = 1;
/**/  int yeshilighting = 1;

/***/ int yeshilighting = 1;
/***/ int yeshilighting = 1;
/***/ int yeshilighting = 1;
/***/ int yeshilighting = 1;
/***/ int yeshilighting = 1;
/**/  int nohilighting  = 1;
/**/  int yeshilighting = 1;
```

Alternatively, just put `/**/` at the beginning of a file.

**Expected behavior:**

Something similar to the above syntax hilighting

**Actual behavior:**

image

**Reproduces how often:**

Always

### Versions
```
Atom    : 1.23.2
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
```

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue raised in #120. When writing something like this:
```java
/**/ int a = 1;
```
Code highlights the entire line as comment. This happens because the entire line is treated as a beginning of javadoc comment that starts with `/**`, and since it is allowed to be multiline, it affects following lines. This situation should be highlighted as empty comment, which we already have in spec. 

Patch uses pattern `'begin': '^\\s*/\\*\\*(?!/)'` for javadoc comment, so we exclude the immediately following `/`. This means that `/**/` is treated as empty comment, `/***/` is treated as javadoc. All other situations `/** /`, `/*/` are unclosed comments and invalid.

### Alternate Designs
None were considered, it looks like this fix is reasonably sound.

### Benefits
Fixes current bug with highlighting empty comments.

### Possible Drawbacks
None that I am aware of.

### Applicable Issues
Fixes #120

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -477,7 +477,7 @@
     ]
   'comments-javadoc':
     'patterns': [
-      'begin': '^\\s*/\\*\\*'
+      'begin': '^\\s*/\\*\\*(?!/)'
       'beginCaptures':
         '0':
           'name': 'punctuation.definition.comment.java'
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -1582,6 +1582,39 @@ describe 'Java grammar', ->
     expect(lines[5][1]).toEqual value: ' javadoc comment ', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'comment.block.javadoc.java']
     expect(lines[5][2]).toEqual value: '*/', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'comment.block.javadoc.java', 'punctuation.definition.comment.java']
 
+  it 'tokenizes empty/single character comment', ->
+    # this test checks the correct tokenizing of empty/single character comments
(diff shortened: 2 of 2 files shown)


sadikovi:
@50Wliu please review. I think this is a more or less reasonable fix - happy to make changes and/or add more tests if needed.

50Wliu:
Looks solid.  I just checked if it would make more sense to order `#comments` and `#comments-javadoc` properly but it looks like nesting of patterns makes that more complicated than this simple fix.

(fyi you can put which issues your PR fixes under "Applicable Issues")

sadikovi:
Thanks @50Wliu! Hopefully this fix does not break other things. 
I forgot about applicable issues, will keep it for future PRs.

PR #122, opened 2018-01-10
Added TODO, FIXME, IDEA snippets

danielct1995:
I think that could be usefull to have snippets of To-do comments like JavaScript


50Wliu:
These should be implemented in [language-todo](https://github.com/atom/language-todo).

Issue #114, opened 2017-10-25
Syntax highlighting is not working with comment


snjeza:
VS Code 1.18

The VS Code master:

![syntax2](https://user-images.githubusercontent.com/388609/32010429-1a1a7088-b9b2-11e7-91de-d627d6872474.png)
The VS Code master with https://github.com/Microsoft/vscode/pull/36786:

![syntax3](https://user-images.githubusercontent.com/388609/32010470-336dae6a-b9b2-11e7-9606-f425faa12ecb.png)

Test2.java
```
package pckg;

public class Test2 
implements Int {
    private JPanel plottingSurface = new JPanel() /* {
        @Override
        public Dimension getPreferredSize() {
            return new Dimension(400, 300);
        }
        @Override
        protected void paintComponent(Graphics g) {
            super.paintComponent(g);
        }
    }*/;

    public static <M extends Map<Integer, String>> 
boolean TestGet(Supplier<M> c) {
 //TODO write unit test for get
 //Setup
 Map<Integer, String> map = c.get();
 return false;
}
}
```
Copied from https://github.com/Microsoft/vscode/pull/36786#issuecomment-339389126

See https://github.com/Microsoft/vscode/issues/36688, https://github.com/redhat-developer/vscode-java/issues/338 and https://github.com/Microsoft/vscode/pull/36786

sadikovi:
Interesting. Several observations:
1. Inline comments work okay, if function definition is on the same line, e.g.
Works:
```java
public static <M extends Map<Integer, String>> boolean TestGet(Supplier<M> c) {
  // works
}
```
Does not work:
```java
public static <M extends Map<Integer, String>> 
boolean TestGet(Supplier<M> c) {
  // does not work
}
```

2. Multi-line comments work if they do not follow `new` keyword.
Works:
```java
private JPanel plottingSurface = /*new JPanel() {
    @Override
    public Dimension getPreferredSize() {
        return new Dimension(400, 300);
    }
    @Override
    protected void paintComponent(Graphics g) {
        super.paintComponent(g);
    }
}*/;
```
Does not work:
```java
private JPanel plottingSurface = new /*JPanel() {
    @Override
    public Dimension getPreferredSize() {
        return new Dimension(400, 300);
    }
    @Override
    protected void paintComponent(Graphics g) {
        super.paintComponent(g);
    }
}*/;
```
I will have a closer look today.

50Wliu:
[...]

PR #123, opened 2018-01-11
Add comments into catch parameters block

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue of missing comments highlighting in `catch` parameters block. For example,
```java
try {
  ...
} catch (IOException /* RuntimeException */ err) {
  ...
}
```

RuntimeException would not be highlighted correctly as comment, this PR fixes it.
The only change is including `#comments` pattern into `meta.catch.parameters.java`.

I added test to make sure that comments in try-catch-finally block are rendered. Note that we still **do not support** something like this:
```java
try {
  ...
} catch /* unsupported comment */ (Exception err) {
  ...
} finally /* unsupported comment */ {
  ...
}
```

I simply do not know how to fix it easily. Would appreciate if someone could help with this (if it needs to be fixed).

### Alternate Designs
None were considered because of the simplicity of the fix.

### Benefits
Fixes comment highlighting in catch parameters block.

### Possible Drawbacks
Might break something else. 

### Applicable Issues
Issue #114 (issue is created for a different bug, but this problem was raised in the comments).

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -588,6 +588,9 @@
             'contentName': 'meta.catch.parameters.java'
             'patterns': [
               {
+                'include': '#comments'
+              }
+              {
                 'include': '#parameters'
               }
             ]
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you review this, please?
There is also a question of supporting comments in catch and finally like I mentioned in PR (unsupported comments section). Would you mind sharing your thought on this? Should we fix it?

Also ping for @tristan957-expero, this should fix the mentioned issue.

50Wliu:
The other comments are a bit harder to catch since you have to change the match itself to recognize comments instead of adding subpatterns.

Fix looks good to me.  Also, since you're a maintainer now, you can create branches directly from this repo rather than from your fork.

tristan957:
Cool thank you for setting my issue in the comments :D

sadikovi:
@50Wliu Thanks for the review!
I will merge it shortly.

PR #124, opened 2018-01-20
Fix variable assignment in try-with-resources pattern

### LINKED ISSUE #88: Resources part of the try-with-resources statement does falsely require a trailing semicolon
### Description

The resources part of the try-with-resources statement does falsely require a trailing semicolon.

### Steps to Reproduce

1. Open new file
2. Enter:
```java
class Main {
    public static void main() {
        try (Foo foo = bar) {
            
        }
    }
}
```

**Expected behavior:**
The `meta.try.resources.java` scope ends at the closing parenthesis and the try body gets the scope `meta.try.body.java`.

**Actual behavior:**
The `meta.try.resources.java` scope does never end and the try body still has the scopes `meta.try.resources.java` and `meta.definition.variable.java`.

**Reproduces how often:**
100%

### Versions
Reproduced in:
```
$atom --version 
Atom    : 1.16.0
Electron: 1.3.13
Chrome  : 52.0.2743.82
Node    : 6.5.0
```
(build from source)

With language-java at  `0.27.0` and in `master` (i.e. 7b4c32a4).

```
$ uname -a
Linux arch-pc 4.10.11-1-ARCH #1 SMP PREEMPT Tue Apr 18 08:39:42 CEST 2017 x86_64 GNU/Linux
```

### Other issues

Support for try-with-resources was originally added in #80 after being reported in #79.

### Additional Information

The Java Spec defines the syntax of this language element as:
```
TryWithResourcesStatement:
    try ResourceSpecification Block [Catches] [Finally]
ResourceSpecification:
    ( ResourceList [;] )
ResourceList:
    Resource {; Resource}
Resource:
    {VariableModifier} UnannType VariableDeclaratorId = Expression
```
See: http://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3

Therefore the semicolon after the "ResourceList" is optional. Furthermore the resources in the try-with-resources can't be arbitrarily code (link it is currently the case), but have to be a special kind of assignment.

Relevant source code: [grammars/java.cson L532](https://github.com/atom/language-java/blob/7b4c32a4c97fa2a01eaaa2ae08a085ee08bba8fb/grammars/java.cson#L532)

I would fix it myself but I have no time at the moment + few experience with atom grammars, but maybe i have some time to look into it in the future.

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue with broken highlighting when resource expression is not closed with `;` in try-with-resources pattern, which is technically allowed. @50Wliu suggested to come up with separate pattern for try-with-resources, which I simplified a bit by keeping original try-catch-finally, but instead creating separate pattern for variable definition. 

Such pattern is called `variables-definition-only` and used in try-with-resources to match the following: `String variable =` or `int num =` without actual expression. We use `#code` in patterns to highlight the expression code. 

In this case `meta.definition.variable.java` only covers type, variable name and `=`. I am not sure if this is required to add code, but it highlights correctly.

After patch (also passes test case mentioned in the issue):
after-patch

### Alternate Designs
Alternative design was considered where we apply a hack to `end` by listing `)` as another terminating character. This is a very small fix (4-8 characters change), but could potentially result in other issues, such as triggering close on closing paren (see https://github.com/atom/language-java/issues/88#issuecomment-333003571).

### Benefits
Fixes issue where not using `;` (which is allowed by specification) would break the highlighting. This change is relatively minor and only affects `try-with-resources`.

### Possible Drawbacks
Since we have to duplicate large regular expression to capture variables (in `variables` and `variables-definition-only`), this could potentially result in updating only one but not the other. I added comment that both patterns should be updated and be in sync, but this might be enough.

### Applicable Issues
Fixes #88

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1237,7 +1237,7 @@
           (?>(\\w+\\.)*[A-Z]+\\w*) # e.g. `javax.ws.rs.Response`, or `String`
         )
         (
-          <[\\w<>,?\\s]*> # HashMap<Integer, String>
+          <[\\w<>,\\.?\\s]*> # e.g. `HashMap<Integer, String>`, or `List<java.lang.String>`
         )?
         (
           (\\[\\])* # int[][]
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -545,7 +545,7 @@
             'name': 'meta.try.resources.java'
             'patterns': [
               {
-                'include': '#variables-definition-only'
+                'include': '#variable-assignment'
               }
               {
                 'include': '#code'
@@ -1207,15 +1207,14 @@
       }
     ]
   # pattern for variable name
-  'variables-name':
+  'variable-name':
     'match': '([A-Za-z$_][\\w$]*)(?=\\s*(\\[\\])*\\s*(;|=|,))'
     'captures':
       '1':
         'name': 'variable.other.definition.java'
-  # pattern for definition of a variable including assignment, but without expression part, e.g.
-  # 'String variable =' or 'int num =', used with '#code' when we cannot rely on ';' to terminate
-  # variable assignment
-  'variables-definition-only':
+  # pattern for a variable assignment (without expression part), e.g. 'String variable =' or
+  # 'int num =', used in places where we cannot rely on ';' as a terminating character for a match
+  'variable-assignment':
     'begin': '''(?x)
       (?=
         (
@@ -1239,20 +1238,21 @@
     'name': 'meta.definition.variable.java'
     'patterns': [
       {
-        'include': '#variables-name'
+        'include': '#variable-name'
       }
       {
         'include': '#all-types'
       }
       {
-        'match': '='
-        'captures':
+        'begin': '='
+        'beginCaptures':
           '0':
             'name': 'keyword.operator.assignment.java'
+        'end': '(?!\\G)'
       }
     ]
-  # pattern for full variable assignment
-  # when updating pattern make sure it is in sync with '#variables-definition-only' pattern
+  # pattern for full variable definition
+  # when updating pattern make sure it is in sync with '#variable-assignment' pattern
   'variables':
     'begin': '''(?x)
       (?=
@@ -1277,7 +1277,7 @@
     'name': 'meta.definition.variable.java'
     'patterns': [
       {
-        'include': '#variables-name'
+        'include': '#variable-name'
       }
       {
         'include': '#all-types'
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -545,6 +545,9 @@
             'name': 'meta.try.resources.java'
             'patterns': [
               {
+                'include': '#variables-definition-only'
+              }
+              {
                 'include': '#code'
               }
             ]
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you review this PR, please? Thanks!

50Wliu:
Will look at both of your PRs next week hopefully.

grammars/java.cson
```diff
@@ -544,6 +544,9 @@
                 'name': 'punctuation.section.try.resources.end.bracket.round.java'
             'name': 'meta.try.resources.java'
             'patterns': [
+              {
+                'include': '#variables-definition-only'
+              }
               {
                 'include': '#code'
```
50Wliu: I don't think this is needed anymore.  Try-with-resources only supports variable definitions.

grammars/java.cson
```diff
@@ -1203,6 +1206,53 @@
         'include': '#object-types'
       }
     ]
+  # pattern for variable name
+  'variables-name':
```
50Wliu: -> `variable-name`

grammars/java.cson
```diff
@@ -1203,6 +1206,53 @@
...
+    'match': '([A-Za-z$_][\\w$]*)(?=\\s*(\\[\\])*\\s*(;|=|,))'
+    'captures':
+      '1':
+        'name': 'variable.other.definition.java'
+  # pattern for definition of a variable including assignment, but without expression part, e.g.
+  # 'String variable =' or 'int num =', used with '#code' when we cannot rely on ';' to terminate
+  # variable assignment
+  'variables-definition-only':
```
50Wliu: What do you think about `variable-assignment` or something?

grammars/java.cson
```diff
@@ -1203,6 +1206,53 @@
...
+    'name': 'meta.definition.variable.java'
+    'patterns': [
+      {
+        'include': '#variables-name'
+      }
+      {
+        'include': '#all-types'
+      }
+      {
+        'match': '='
```
50Wliu: Have you considered making this a begin/end like `variables` but ending it on `(?!\\G)` instead?  I'm curious if that will work or not.

grammars/java.cson
sadikovi: 👍

grammars/java.cson
sadikovi: 👍

grammars/java.cson
sadikovi: I have not considered it, but I updated the code and it works (assuming that I updated it with what you had in mind).

grammars/java.cson
sadikovi: I tried removing `#code`, but it affects syntax highlighting after `=`, for example, in `try (Buffer a = new Buffer())`, `new Buffer()` part will not be highlighted. So I decided to keep it.

sadikovi:
@50Wliu I addressed your comments. One thing that I decided not to do is removing `#code` from try with resources, because it does not highlight part after `=`. But it could also mean that I missed something - would appreciate if you could review again.

Also I changed match to begin/end, and it worked. Could you review that part to make sure that I did it right? I do not quite understand the `?!\\G` pattern.

Thanks!

50Wliu:
`\G` means "where the last match ended".  So `\Ga` means "match `a` only if it immediately follows the last match".  Therefore, its negation `(?!\G)` means something akin to "start matching only if we are not immediately following where the previous match ended".  So doing `'end': '(?!\\G)'` then means "end the match when there are no more continuous matches".

You'll have to test something like `var = 3 + 2` to make sure it works correctly though, because I still have trouble deciding when to use `\G`.

grammars/java.cson
```diff
@@ -1203,6 +1206,53 @@
...
+      }
+      {
+        'include': '#all-types'
+      }
+      {
+        'begin': '='
+        'beginCaptures':
+          '0':
+            'name': 'keyword.operator.assignment.java'
+        'end': '(?!\\G)'
```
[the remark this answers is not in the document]
sadikovi: I tried it, updated code as follows:
```diff
diff --git a/grammars/java.cson b/grammars/java.cson
index c6c513e..6700101 100644
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -547,9 +547,6 @@
               {
                 'include': '#variable-assignment'
               }
-              {
-                'include': '#code'
-              }
             ]
           }
           {
@@ -1249,6 +1246,11 @@
           '0':
             'name': 'keyword.operator.assignment.java'
         'end': '(?!\\G)'
+        'patterns': [
+          {
+            'include': '#code'
+          }
+        ]
       }
     ]
   # pattern for full variable definition
```

Did not work. I also tried couple of different combinations - same result.

sadikovi:
@50Wliu I updated the code. Have you decided if we should go with `\G`? What do you think about overall approach? Do you want me to make any other changes? Thanks.

sadikovi:
@50Wliu Is this change still relevant? If so, I suggest we deal with other opened PRs, so I can rebase this one and update if necessary (because of some changes in other pull requests).

50Wliu:
Since this change is a little larger I'll be reviewing it separately within the coming days.

sadikovi:
@50Wliu I rebased this PR. Could you review it again, when you have time? Thanks!
Do you have any concerns with this work? One thing I can think of is duplicated code for `variable-assignment` and `variables`, but I do not know how to make it reusable.

sadikovi:
This needs additional reviewing and potentially different fix, since we have to duplicate patterns here. I will have a look again, see if there is a better approach.

sadikovi:
I am going to close this PR, since it introduces a lot of unnecessary changes. I will open an alternative solution shortly, so we can choose the one to go with.

PR #125, opened 2018-01-21
Fix unified expression language single quote escape

### LINKED ISSUE #83: Syntax highlighting confused by JSTL replace pattern
### Description

The syntax highlighting for JavaServer Pages is being thrown off by a JSTL string replace pattern that consists of an escaped single quote:

`
<c:set var="foo" value="${fn:replace(entry.name, '\\'', '&#39;')}" />
<c:set var="bar" value="test" />
`

The problem is with the `'\\''` segment. 

### Steps to Reproduce

1. Open a new editor tab and set the language to JavaServer Pages
2. Paste in the line above

The syntax highlighting rules should catch the mismatched single quotes and ignore them.

The odd number of single quotes confuses the syntax highlighter, and it treats any successive lines in the file as being string values.

100%

### Versions

Atom 1.15.0 x64
Running on OS X 10.11.6

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue when escaping single quote in single quoted string affects subsequent highlighting, because current pattern matches the first single quote as the end of the literal, even though it is escaped. It was realised that we only need to match ending of literal when it is not escaped, any single quote with preceding '\\' should be marked as escaped.

Before patch:
before-patch

After patch:
after-patch

Also correctly highlights situation `'\\   '`, though I am not sure if this is a correct syntax. 

### Alternate Designs
None were considered apart from the PR solution.

### Benefits
Fixes issue of highlighting when escaping single quote in single quote string literal.

### Possible Drawbacks
Might break some other code, because I am not very familiar with expression language and could have overlooked some escape situations, such as using the `\\'` as a value.

### Applicable Issues
Fixes #83

### DIFF
--- a/grammars/unified expression language (el).cson
+++ b/grammars/unified expression language (el).cson
@@ -57,7 +57,7 @@
     'beginCaptures':
       '0':
         'name': 'punctuation.definition.string.begin.java.el'
-    'end': '\''
+    'end': '(?<!\\\\)\''
     'endCaptures':
       '0':
         'name': 'punctuation.definition.string.end.java.el'
--- a/spec/unified-el-spec.coffee
+++ b/spec/unified-el-spec.coffee
@@ -157,6 +157,14 @@ describe 'Unified expression language grammar', ->
       expect(tokens[5]).toEqual value: 'baz', scopes: ['source.java.el', 'string.quoted.single.java.el']
       expect(tokens[6]).toEqual value: "'", scopes: ['source.java.el', 'string.quoted.single.java.el', 'punctuation.definition.string.end.java.el']
 
+    it 'tokenizes single quoted string literals with escaped single quote', ->
+      {tokens} = grammar.tokenizeLine "'foo \\\\\' bar'"
+      expect(tokens[0]).toEqual value: "'", scopes: ['source.java.el', 'string.quoted.single.java.el', 'punctuation.definition.string.begin.java.el']
+      expect(tokens[1]).toEqual value: 'foo ', scopes: ['source.java.el', 'string.quoted.single.java.el']
+      expect(tokens[2]).toEqual value: '\\\\', scopes: ['source.java.el', 'string.quoted.single.java.el', 'constant.character.escape.java.el']
+      expect(tokens[3]).toEqual value: '\' bar', scopes: ['source.java.el', 'string.quoted.single.java.el']
+      expect(tokens[4]).toEqual value: "'", scopes: ['source.java.el', 'string.quoted.single.java.el', 'punctuation.definition.string.end.java.el']
+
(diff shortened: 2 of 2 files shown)


sadikovi:
@50Wliu could you review this pull request? Thanks!

50Wliu:
@MoritzKn May I ask for your input here?  Do you need to double-escape backslashes in EL?

MoritzKn:
@50Wliu I haven't used it for quite some time and don't have a setup at hand to test it right now. The specification says:
> 1.3 Literals
> There are literals for boolean, integer, floating point, string, and null in an evalexpression.
> - Boolean - true and false
> - Integer - As defined by the IntegerLiteral construct in Section 1.19
> - Floating point - As defined by the FloatingPointLiteral construct in
Section 1.19
> - String - With single and double quotes - " is escaped as \", ' is escaped as \',
and \ is escaped as \\. Quotes only need to be escaped in a string value enclosed
in the same type of quote
> - Null - null

-- http://download.oracle.com/otn-pub/jcp/jsp-2.1-fr-spec-oth-JSpec/jsp-2_1-fr-spec-el.pdf

However this may  not apply for this case since it's embedded in the JSP code.

50Wliu:
My hunch is that _since_ this is embedded in JSP, we should be adding escape character support to JSP instead of EL.

sadikovi:
The reason I fixed in EL was due to broken `punctuation.definition.string.end.java.el` being defined in unified expression language, so I thought that it must be the place to fix it, rather than JSP, which does not define it. Another argument to support this was that bug originates from a function being part of the attribute value.

My understanding is that JSP file is for tags and attributes (key = value), and values are parsed as EL (scope `source.java.el`) - though I might be wrong here, and would appreciate any corrections. Thanks!

50Wliu:
From the "Quoting and Escape Conventions" section of http://download.oracle.com/otn-pub/jcp/jsp-2_3-mrel2-eval-spec/JSP2.3MR.pdf?AuthParam=1517623843_d412f1e03b76e1d552dbf301b0948086 (not sure how recent that is), backslash escaping is done in attribute values (in this case EL).  So I think what's happening here is the same thing as what's happening when writing grammar files - the backslash needs to be double-escaped once for EL and then the second time for JSP.

sadikovi:
@50Wliu I think this is what we do - escape only `\\'`, right? Should we proceed with this fix or do you have any concerns? Thanks.

50Wliu:
@sadikovi No, I believe that we should add proper escape tokenization to the JSP language.

sadikovi:
@50Wliu thanks for the review and comment! I used previous escape scope as a guidance and thought that this fox should be there as well. I am going to close this PR - currently there are other more critical issues to fix:)

PR #128, opened 2018-02-17
Fix capitalized variables being highlighted as storage type

### LINKED ISSUE #76: Capital variables tokenized as storage types

sadikovi:
### Requirements

### Description of the Change
This PR fixes highlighting of variables that are capitalized or start with capital letter. I fixed by enforcing that the last branch of 'object-types' pattern requires the next word (variable name); and since variable names can start with `A-Za-z` or `_`, or `$`, I added `\w|$`.

Tested on the code:
```java
class A {
  void func() {
    int g = 1;
    g += 2;
    int G = 1;
    G += 2;
    ABC += 3;
    Abc += 4;

    if (G > g) {
      // do something
      String a = "abc";
    }
  }
}
```

### Alternate Designs
None were considered, though I was experimenting with different changes to see what actually affects the highlighting. 

### Benefits
Fix problem with highlighting variables starting with capital letters in situations like, but not limited to:
```java
A += 2;
A = 3;
if (A > B) {
  // something
}
```

### Possible Drawbacks
Might break something else, but it passes all tests so far.

### Applicable Issues
Closes #76

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1003,8 +1003,9 @@
         ]
       }
       {
-        # If the above fails *then* just look for Wow (must be followed by a variable name)
-        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b(?=\\s*(\\w|\\$))'
+        # If the above fails *then* just look for Wow
+        # (must be followed by a variable name, we use '\n' to cover multi-line definitions)
+        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b(?=\\s*[A-Za-z$_\\n])'
         'name': 'storage.type.java'
         'captures':
           '1':
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1003,8 +1003,8 @@
         ]
       }
       {
-        # If the above fails *then* just look for Wow
-        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b'
+        # If the above fails *then* just look for Wow (must be followed by a variable name)
+        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b(?=\\s*(\\w|\\$))'
         'name': 'storage.type.java'
         'captures':
           '1':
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you review this PR, please? Thanks!

grammars/java.cson
```diff
@@ -1003,8 +1003,8 @@
         ]
       }
       {
-        # If the above fails *then* just look for Wow
-        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b'
+        # If the above fails *then* just look for Wow (must be followed by a variable name)
+        'match': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*\\b(?=\\s*(\\w|\\$))'
```
50Wliu: Nitpick: `[\\w$]` might be clearer and is typically what we do with languages that accept `$` as an identifier.  Also I think variables can start with underscores?

50Wliu:
My only concern: I think Java allows multiline variable declarations like
```java
String
test = 3;
```

Please tell me if I'm wrong.  Even if that's allowed, I'm thinking that this change should be merged because uppercase variables are more likely than multiline declarations.

grammars/java.cson
sadikovi: Yes, you are right. I actually replaced with the following: `[A-Za-z$_\\n]`, which is what we use for start of the variable, and I added new line to cover cases:
```
String
test = "abc";
```

With this change, current highlighting is the same as in master branch (which is not exactly the same as inline definition `String test = "abc"`).

sadikovi:
I agree. We should push this effort instead. I also updated the line that you commented on. Now highlighting is similar to the master branch for multi-line definitions. See images below:

### Master branch:
master

### This patch
branch

@50Wliu Could you review this again? Thanks!

PR #133, opened 2018-03-05
Fix highlighting for generics with dots in class name

### LINKED ISSUE #131: Problems with highlighting generics when class name has dots, e.g. Set<Map.Entry<K, V>>
### Description
This is an issue reported as part of #127.

Reproducible in the latest master 7a42895fead16e2c39c6667fc9e7ce0f7f922ac4.

Code to reproduce:
```java
class A {
  Set<Map.Entry<K, V>> something = null;
  Set<MapEntry<K, V>> something = null;
}
```

Screenshot with current highlighting - as you see, generics with dots are scoped differently.

screenshot from 2018-02-21 09-35-54

### Steps to Reproduce

1. Open Atom new tab
2. Select Java language highlighting
3. Paste code snippet above
4. Result should be exactly the same as on the image

Expected behaviour is that both lines are highlighted in the same way, should be similar to highlighting of the generics that do not have dots in the class names.

Both lines are highlighted differently.

Reproduced all the time on current master branch.

### Versions
I have tested this on OS X and Ubuntu - reproducible on both. 
Atom 1.23.3+.
`language-java` is the latest master.

### Additional Information
This was reported originally in #127. For more information, please see that issue.

sadikovi:
### Requirements

### Description of the Change
This PR fixes issue when having code like `Set<java.lang.String> var = ...`, `java.lang.String` part was highlighted incorrectly. 

Consider code below:
```java
class A {
  Set<MapEntry<K, V>> var = null;
  Set<Map.Entry<K, V>> varA = null;
  Set<java.util.List<K>> varB = null;
}
```

#### Before the patch:
before

#### After the patch
after

### Alternate Designs
None were considered. I just tried different combinations of patterns to see which one worked.

### Benefits
Fixes the issue, without breaking existing tests.

### Possible Drawbacks
This does not fix code like following, parts `java.util.Set` and `java.lang.String` are highlighted incorrectly:
```java
class A {
  java.util.Set<java.util.List<K>> varA = null;
  java.lang.String varB = null;
}
```
But I think this is outside of the scope of this PR.

### Applicable Issues
Fixes #131

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -697,6 +697,10 @@
         'name': 'punctuation.separator.delimiter.java'
       }
       {
+        'match': '\\.'
+        'name': 'punctuation.separator.period.java'
+      }
+      {
         'include': '#parens'
       }
       {
@@ -1216,7 +1220,7 @@
           (?>(\\w+\\.)*[A-Z]+\\w*) # e.g. `javax.ws.rs.Response`, or `String`
         )
         (
-          <[\\w<>,?\\s]*> # HashMap<Integer, String>
+          <[\\w<>,\\.?\\s]*> # e.g. `HashMap<Integer, String>`, or `List<java.lang.String>`
         )?
         (
           (\\[\\])* # int[][]
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -672,7 +672,7 @@
         'name': 'storage.modifier.$1.java'
       }
       {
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu Could you please review this PR? Thanks!
Let me know what you think about **Possible Drawbacks** section. Should we fix it here or create a separate issue? Thanks.

50Wliu:
Separate issue is fine.

PR #134, opened 2018-03-06
Fix master branch: add varargs into object-types pattern

sadikovi:
### Requirements

### Description of the Change
This PR fixes master branch. It was broken after I merged varargs -> capitalised variables pull requests.
The reason it started failing is that for capital variables we check if storage type is followed by variable name. This does not work for `...` varargs, e.g. `String ... args`. So I included it in the regexp. Yes, it is not pretty, but fixes tests and I checked manually with different codes.

### Alternate Designs
None, unfortunately, just wanted to fix it. Other ways would be reconsidering how to work with capital variables.

### Benefits
Fixes master branch.

### Possible Drawbacks
The fix makes `object-types` scope messy.

### Applicable Issues
None

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1008,8 +1008,9 @@
       }
       {
         # If the above fails *then* just look for Wow
-        # (must be followed by a variable name, we use '\n' to cover multi-line definitions)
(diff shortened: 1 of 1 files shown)


sadikovi:
@50Wliu I "patched" the code to pass tests. Could you review this, please? 
Sorry for breaking it in the first place.

50Wliu:
> Sorry for breaking it in the first place.

No problem, just wait for tests to pass before merging next time :).

Issue #136, opened 2018-03-12
Missing keywords


SMadani:
The following reserved keywords are missing from the Java grammar:

- const
- goto

Furthermore, Java 10 adds local-variable type inference, so perhaps var should also be added in certain contexts?

I have been referred to this project from Visual Studio Code. The original problem can be found [here.](https://github.com/Microsoft/vscode/issues/45405)

50Wliu:
From https://docs.oracle.com/javase/tutorial/java/nutsandbolts/_keywords.html, it appears that `const` and `goto` are unused as of Java 8.  Has that changed since then?  If not, I don't see a reason to add them.

SMadani:
@50Wliu Although they serve no purpose, they are still reserved keywords so an editor should be aware of their status as a keyword rather than, for example, allowing a user who is unfamiliar with the language to use them as variable names and get compilation errors.

50Wliu:
That makes sense to me, thanks for the explanation.  Would you like to see them tokenized as regular keywords or for some sort of warning highlighting to be applied? /cc @sadikovi

SMadani:
Thank you for considering this. I think it would make sense to treat them as regular keywords as nearly all other editors do not distinguish between "active" and "inactive" reserved keywords, though a warning would also be a bonus.

sadikovi:
I think it is worth adding `goto` and `const` as regular keywords, but I do not think we should be adding `var`. If everyone agrees, I can open a PR for this.

SMadani:
Great, thanks!

fbricon:
How is `var` supposed to be handled then? It's technically not a keyword but a reserved type name, but we still need it to be coloured differently:

 
<img width="301" alt="screen shot 2018-04-01 at 6 43 05 pm" src="https://user-images.githubusercontent.com/148698/38178247-a654ca8e-35dc-11e8-8d43-f8b3c3ba11bb.png">

SMadani:
It would probably make sense to colour it as a regular keyword in places where it actually works as a keyword, such as the example you provided. In general, this applies to variable declarations in methods, for loop initializers, try-with-resources and from Java 11, in lambda parameter declarations. However this is probably too complicated to support at an editor level.

sadikovi:
As you pointed out, `var` is neither a reserved keyword nor a keyword at all, and could be used perfectly for a variable name, therefore we do not apply any special highlighting. We could always try adding support for `var` reserved type name, but not as part of this issue.

[...]

PR #137, opened 2018-03-13
Add reserved keywords

sadikovi:
### Requirements

### Description of the Change
This PR adds reserved keywords `const` and `goto`, so they are highlighted as standard keywords, e.g. `try` or `catch`, or `instanceof`. I assigned scope `keyword.reserved.java` for them, because I could not figure out what existing scope they should belong to, since they are not used:).

Looks like this now:
after

### Alternate Designs
None were considered, because it is fairly simple change.

### Benefits
Reserved keywords are now highlighted the same way as standard keywords.

### Possible Drawbacks
None.

### Applicable Issues
#136

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -814,6 +814,10 @@
         'match': '(\\||&)'
         'name': 'keyword.operator.bitwise.java'
       }
+      {
+        'match': '\\b(const|goto)\\b'
+        'name': 'keyword.reserved.java'
+      }
     ]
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu Could you review this PR? Let me know if it should be added like this or we should add those keywords into one of the existing scopes. Thanks!

sadikovi:
Thanks @50Wliu! Will merge.

PR #143, opened 2018-05-29
Add Java 9/10 keywords

### LINKED ISSUE #142: Add Java 9/10 keywords
Java 9 introduces the following keywords:
open,module,requires,transitive,exports,opens,to,uses,provides and with

See §3.9 in http://cr.openjdk.java.net/~mr/jigsaw/spec/java-se-9-jls-diffs.pdf

It will enable module-info.java syntax highlighting.
See https://github.com/redhat-developer/vscode-java/issues/527

snjeza:
## Requirements

- Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
- All new code requires tests to ensure against regressions

## Description of the Change

This PR adds reserved keywords open,module,requires,transitive,exports,opens,to,uses,provides and with, so they are highlighted as standard keywords. I assigned scope keyword.modules.java for them, because they are used in modules-info.java.
See §3.9 in http://cr.openjdk.java.net/~mr/jigsaw/spec/java-se-9-jls-diffs.pdf

Before:
527a

After:

## Alternate Designs

None were considered, because it is fairly simple change.

## Benefits

Java 9/10 keywords are now highlighted the same way as standard keywords.

## Possible Drawbacks

None.

## Applicable Issues

#142

### DIFF
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -192,6 +192,22 @@ describe 'Java grammar', ->
 
     expect(tokens[9]).toEqual value: 'goto', scopes: ['source.java', 'keyword.reserved.java']
 
+  it 'tokenizes module keywords', ->
+    lines = grammar.tokenizeLines '''
+      module Provider {
+        requires ServiceInterface;
+        provides javax0.serviceinterface.ServiceInterface with javax0.serviceprovider.Provider;
+      }
+    '''
+
+    expect(lines[0][0]).toEqual value: 'module', scopes: ['source.java', 'meta.module.java', 'storage.modifier.java']
+    expect(lines[0][2]).toEqual value: 'Provider', scopes: ['source.java', 'meta.module.java', 'entity.name.type.module.java']
+    expect(lines[0][4]).toEqual value: '{', scopes: ['source.java', 'meta.module.java', 'punctuation.section.module.begin.bracket.curly.java']
+    expect(lines[1][1]).toEqual value: 'requires', scopes: ['source.java', 'meta.module.java', 'meta.module.body.java', 'keyword.module.java']
+    expect(lines[2][1]).toEqual value: 'provides', scopes: ['source.java', 'meta.module.java', 'meta.module.body.java', 'keyword.module.java']
+    expect(lines[2][3]).toEqual value: 'with', scopes: ['source.java', 'meta.module.java', 'meta.module.body.java', 'keyword.module.java']
+    expect(lines[3][0]).toEqual value: '}', scopes: ['source.java', 'meta.module.java', 'punctuation.section.module.end.bracket.curly.java']
+
   it 'tokenizes classes', ->
     lines = grammar.tokenizeLines '''
       class Thing {
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -109,6 +109,9 @@
   {
     'include': '#code'
   }
+  {
+    'include': '#module'
+  }
 ]
 'repository':
   'all-types':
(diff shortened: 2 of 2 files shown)


snjeza:
module-info.java before the patch:

sadikovi:
Thank you for the pull request, I will have a look shortly.
Could you fill out pull request template that describes the change and tests? Thanks.

sadikovi:
Looks good, but these keywords look specific to a module. For example, I was thinking if it would be better to have `module` scope similar to `class` scope, since they are very similar. And have all those keywords within the module scope. This approach would not break existing highlighting in the code that uses them as identifiers - the keywords are only "keywords" within module, not in other code.

Here is an example (one of the ways to do it):

Include our scope to global `patterns`, since it has its own file, sort of:
```cson
{
  'include': '#module'
}
```

Then we would create an initial draft of module, which is already pretty good. This allows us to extend it later with other rules, e.g. `requires XYZ;` only on one line, etc.
```cson
'module':
  # a uniquely named, reusable group of related packages, as well as resources (such as images
  # and XML files).
  'begin': '((open)\\s)?(module)\\s+(\\w+)'
  'end': '}'
  'beginCaptures':
    '1':
      'name': 'storage.modifier.java'
    '3':
      'name': 'storage.modifier.java'
    '4':
      'name': 'entity.name.type.module.java'
  'endCaptures':
    '0':
      'name': 'punctuation.section.module.end.bracket.curly.java'
  'name': 'meta.module.java'
  'patterns': [
    {
      'begin': '{'
      'beginCaptures':
        '0':
          'name': 'punctuation.section.module.begin.bracket.curly.java'
      'end': '(?=})'
      'contentName': 'meta.module.body.java'
      'patterns': [
        # TODO: Write more rules for module grammar
        {
          'match': '\\b(requires|transitive|exports|opens|to|uses|provides|with)\\b'
          'name': 'keyword.module.java'
        }
      ]
    }
  ]
```

Below is the diff for this, you can use it if you want, or you can write your own (it is a fun exercise, IMHO):

https://gist.github.com/sadikovi/d1a400ef3020be2e07999f20d47fbe74

sadikovi:
@50Wliu could you let us what you think about the module keywords? Thanks.

snjeza:
@sadikovi I have updated the PR.

sadikovi:
Thanks! This change will allow anyone to easily update/improve module later without breaking other code. Could you add a unit test that verifies the change. You can find the examples in `spec/java-spec.coffee`.

sadikovi:
@snjeza Looks good! Could you also check `Provider`, `{` and `}` scopes as well in the added test?

sadikovi:
Looks good. @50Wliu do you have any more comments? Otherwise, I will merge it!

spec/java-spec.coffee
```diff
@@ -192,6 +192,22 @@ describe 'Java grammar', ->
...
 
+  it 'tokenizes module keywords', ->
+    lines = grammar.tokenizeLines '''
+      module Provider {
+        requires ServiceInterface;
+        provides javax0.serviceinterface.ServiceInterface with javax0.serviceprovider.Provider;
+      }
+    '''
+
+    expect(lines[0][0]).toEqual value: 'module', scopes: [ 'source.java', 'meta.module.java', 'storage.modifier.java' ]
```
50Wliu: For consistency, there shouldn't be space padding on the insides of the brackets.

sadikovi:
@snjeza Thanks the PR! @50Wliu Thanks for the review!

PR #144, opened 2018-05-30
Fix meta.catch.java and meta.throwables.java scopes

### LINKED ISSUE #138: Missing scopes on comma and xor
Comma and xor operator are not scoped between exceptions:

```java
public class MyClass {
    public void set(String str) throws A, B { /* , not scoped */
        try {
            /* Something that throws C or D */
        } catch (C | D e) { /* | not scoped */
            throw new A();
        }
    }
}
```

language-java 0.29.0
Atom 1.26.1

sadikovi:
### Requirements

### Description of the Change

This PR updates the code for scopes `meta.catch.java` and `meta.throwables.java` and fixes scoping of exception class names and/or parameters.

For `meta.catch.java` we are currently include `parameters` scope, which is not exactly correct, because parameters are assumed to be `storage + variable`, but in the case of catch block, you could see something like this `Exception1 | Exception2 err`, so we would only highlight the second case. I replaced it with the pattern that scopes `|` as separator and require `storage` and optional `variable` name.

For `meta.throwables.java` we are currently including `object-types` scope, which is also not 100% correct, because they assume `storage` followed by `variable`, but in `throws` you could see something like this `throws Exception1, Exception2`. I changed it to just parse storage type.

```java
class MyClass {
  public void set(String str) throws Exception3, Exception4 {
    try {
      // some code
    } catch (Exception1 | Exception2 e) {
      throw new Exception3();
    }
  }
}
```

Before:
before

After:
after

### Alternate Designs
I was considering updating `parameters` and `object-types` scopes, but I found that it would be rather massive change, which might cause the regression in other parts. So I went with this fairly straightforward approach.

### Benefits
Fixes issue of scoping exceptions in `catch` and `throws`.

### Possible Drawbacks
None that I found, it does not affect existing scopes.

### Applicable Issues
Closes #138

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -595,7 +595,16 @@
                 'include': '#comments'
               }
               {
-                'include': '#parameters'
+                'match': '\\|'
+                'name': 'punctuation.catch.separator'
+              }
+              {
+                'match': '([a-zA-Z$_][\\.a-zA-Z0-9$_]*)\\s*(\\w+)*'
+                'captures':
+                  '1':
+                    'name': 'storage.type.java'
+                  '2':
+                    'name': 'variable.parameter.java'
               }
             ]
           }
@@ -1218,7 +1227,12 @@
     'name': 'meta.throwables.java'
     'patterns': [
       {
-        'include': '#object-types'
+        'match': ','
+        'name': 'punctuation.throwables.separator'
+      }
+      {
+        'match': '[a-zA-Z$_][\\.a-zA-Z0-9$_]*'
+        'name': 'storage.type.java'
       }
     ]
   'variables':
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -596,10 +596,10 @@
               }
               {
                 'match': '\\|'
-                'name': 'punctuation.catch.separator'
+                'name': 'punctuation.catch.separator.java'
               }
               {
-                'match': '([a-zA-Z$_][\\.a-zA-Z0-9$_]*)\\s*(\\w+)*'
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you have a look at this PR? Thanks!

grammars/java.cson
```diff
@@ -595,7 +595,16 @@
                 'include': '#comments'
               }
               {
-                'include': '#parameters'
+                'match': '\\|'
+                'name': 'punctuation.catch.separator'
```
50Wliu: .java

grammars/java.cson
```diff
@@ -595,7 +595,16 @@
...
+              }
+              {
+                'match': '([a-zA-Z$_][\\.a-zA-Z0-9$_]*)\\s*(\\w+)*'
```
50Wliu: Should it be `(\\w+)?` instead?

grammars/java.cson
```diff
@@ -1218,7 +1227,12 @@
     'name': 'meta.throwables.java'
     'patterns': [
       {
-        'include': '#object-types'
+        'match': ','
+        'name': 'punctuation.throwables.separator'
```
50Wliu: .java

I also think this can use the standard comma separator scope.

grammars/java.cson
sadikovi: Yes, you are right. I will update the code.

grammars/java.cson
sadikovi: Yes, you are right. Will update.

sadikovi:
@50Wliu thanks for review! I updated the code, would you mind taking a look again? Thanks.

PR #145, opened 2018-05-30
Fix class fields with arrays as generic type

### LINKED ISSUE #139: Missing scopes on new and punctuation
`=`, `new` and `()` aren't highlighted in the example below:

```java
public class MyClass {
    private ArrayList<int[]> sadFace = new ArrayList<int[]>(); /* doesn't work */
    private ArrayList<Integer> happyFace = new ArrayList<Integer>(); /* works */
}
```

language-java 0.29.0
Atom 1.26.1

sadikovi:
### Requirements

### Description of the Change

This PR fixes the scope of classes with generic type being an array, e.g. `Foo<int[]>`. It scopes properly with the actual assignment, but variable declaration was broken (IMHO, we just forgot to add it).

Fix is simply adding missing `[` and `]` to the pattern. It looked like when I added a test case I was using the wrong scopes, so I changed them to actual correct ones.

Before:
before

After:
after

### Alternate Designs
None were considered, due to relatively simple change.

### Benefits
Fixes a bug with scoping generics with arrays.

### Possible Drawbacks
None were found.

### Applicable Issues
Closes #139

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1230,7 +1230,7 @@
           (?>(\\w+\\.)*[A-Z]+\\w*) # e.g. `javax.ws.rs.Response`, or `String`
         )
         (
-          <[\\w<>,\\.?\\s]*> # e.g. `HashMap<Integer, String>`, or `List<java.lang.String>`
+          <[\\w<>,\\.?\\s\\[\\]]*> # e.g. `HashMap<Integer, String>`, or `List<java.lang.String>`
         )?
         (
           (\\[\\])* # int[][]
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you have a look at this PR? Thanks!

sadikovi:
I also tested the fix manually on different examples of initialising variables, so far this change has been enough to provide correct scope and highlighting.

sadikovi:
I do not think we need to change in other places, since manual tests worked and unit tests pass. I suggest we address any other issues in follow-up PRs.

50Wliu:
Sounds good.

PR #146, opened 2018-07-03
Update 'variables' and 'member-variables' patterns to handle semicolon behaviour

### LINKED ISSUE #88: Resources part of the try-with-resources statement does falsely require a trailing semicolon

sadikovi:
### Requirements

### Description of the Change

This PR updates `variables` and `member-variables` patterns and changes the behaviour of the `meta.definition.variable.java` scope. This is done to fix #88 (try-with-resources issue) to handle `;`.

I propose to define `meta.definition.variable.java` to be just declaration part without the following expression to initialise the variable, e.g. `[int a] = 5;` or `[int a];`, where `[...]` is `meta.definition.variable.java`. This results in the right hand side not having the scope, but I could potentially define and assign another one, like `meta.initialization.expression.java` or something similar.

After patch (also passes test case mentioned in the issue):
```java
class A {
  void func() {
    try (Buffer a = new Buffer()) {
      str = "str";
    }

    try (
      Buffer a = new Buffer();
      Buffer b = new Buffer();
      Buffer c = new Buffer()
    ) {
      str = "str";
    }
  }
}
```

after-patch

I updated the existing tests to remove `meta.definition.variable.java` wherever it is not needed, and added more tests to check the correct behaviour.

### Alternate Designs
Alternative design was considered where we apply a hack to `end` by listing `)` as another terminating character. This is a very small fix (4-8 characters change), but could potentially result in other issues, such as triggering close on closing paren (see https://github.com/atom/language-java/issues/88#issuecomment-333003571).

Another alternative design was proposed in #124, but it is more intrusive and results in duplicating variables pattern code and increases maintenance cost, since both patterns must be updated in case of any changes.

### Benefits
Fixes issue when not using `;` (which is allowed by specification) would break the highlighting.
Potentially helps to fix other problems related to variable highlighting.

### Possible Drawbacks
We remove `meta.definition.variable.java` scope from the expression that follows `=`. The left hand side is okay. Arguably, the right hand side should not be scoped as variable definition.

### Applicable Issues
Fixes #88
Relates to PR #124

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1309,7 +1309,7 @@
     ]
   'member-variables':
     'begin': '(?=private|protected|public|native|synchronized|abstract|threadsafe|transient|static|final)'
-    'end': '(?=;)'
+    'end': '(?=\\=|;)'
     'patterns': [
       {
         'include': '#storage-modifiers'
@@ -1322,13 +1322,5 @@
       }
       {
         'include': '#object-types'
-      }
-      {
-        'begin': '(?=\\=)'
-        'patterns': [
-          {
-            'include': '#code'
-          }
-        ]
       }
     ]
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1291,7 +1291,7 @@
         \\s*(=|;)
       )
     '''
-    'end': '(?=;)'
+    'end': '(?=\\=|;)'
     'name': 'meta.definition.variable.java'
     'patterns': [
       {
@@ -1302,18 +1302,6 @@
       }
       {
         'include': '#all-types'
-      }
-      {
-        'begin': '='
-        'beginCaptures':
-          '0':
-            'name': 'keyword.operator.assignment.java'
-        'end': '(?=;)'
-        'patterns': [
-          {
-            'include': '#code'
-          }
-        ]
       }
       {
         'include': '#code'
@@ -1334,5 +1322,13 @@
       }
       {
         'include': '#object-types'
+      }
+      {
+        'begin': '(?=\\=)'
+        'patterns': [
+          {
+            'include': '#code'
+          }
+        ]
       }
     ]
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -1606,6 +1606,7 @@ describe 'Java grammar', ->
         private void fn() {
           try (
             BufferedReader in = new BufferedReader();
+            BufferedReader br = new BufferedReader(new FileReader(path))
           ) {
             // stuff
           }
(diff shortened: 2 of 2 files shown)


sadikovi:
@50Wliu can you have a look at this PR? It is my another attempt to fix try-with-resources issue. Thanks!

50Wliu:
Without looking at the code changes yet, this seems like a very reasonable approach :).

grammars/java.cson
```diff
@@ -1335,4 +1323,12 @@
       {
         'include': '#object-types'
       }
+      {
+        'begin': '(?=\\=)'
```
50Wliu: Instead of having to do this, could we also stop `member-variables` at `=`?

50Wliu:
> I could potentially define and assign another one, like meta.initialization.expression.java or something similar.

This sounds good, but can be done in a follow-up PR.

sadikovi:
@50Wliu I updated the PR, let me know if this is what you had in mind. 

I was thinking, do we actually need to add a scope for the right hand side (assignment part) of the variables? Could it be just `#code`?

sadikovi:
@50Wliu Thanks for the review. I am going to merge this if there are no objections.

sadikovi:
I squashed and merged the PR, so we can easily revert a single commit, if needed.

Issue #135, opened 2018-03-06
Incorrect highlighting for generics with storage type containing dots


sadikovi:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description
After we fixed #133, we still left the case when storage type that precedes generic type contains dots in the name, e.g. `java.util.List<String> var` will highlight `java.util.List` incorrectly.

See example below, parts java.util.Set and java.lang.String are highlighted incorrectly:
```java
class A {
  // incorrect examples
  java.util.Set<java.util.List<K>> varA = null;
  java.lang.String varB = null;

  // correct examples
  Set<java.util.List<K>> varC = null;
  String varD = null; 
}
```

![screenshot from 2018-03-06 13-12-07](https://user-images.githubusercontent.com/7788766/37006991-07c1aa0a-2140-11e8-92b1-bdca5bb71ec6.png)

### Steps to Reproduce

1. Open Atom new tab
2. Choose Java language for the tab
3. Paste the code above
4. Highlighting is different 

**Expected behavior:** [What you expect to happen]
Correct highlighting similar to `// correct examples` section in the code

**Actual behavior:** [What actually happens]
Examples under `// incorrect examples` are highlighted incorrectly.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions
OS X and Ubuntu.

apm  1.18.12
npm  3.10.10
node 6.9.5 x64
atom 1.23.3
python 2.7.12
git 2.7.4

Latest master branch (c63382b0706cfe695f778779b77d5c806b47d762).

### Additional Information

None, should be simple to reproduce.

sadikovi:
Fixed in #147.

PR #147, opened 2018-07-17
Fix highlighting for fully qualified types

brennenberkley:
### Description of the Change

Fixes issue where types containing periods were being treated as objects and properties, so I included the ```variables``` section before the ```object``` and ```property``` sections to properly recognize fully qualified variables. I also changed the regex for recognizing types to allow namespaces starting with a lowercase letter.

I also added a few specs. 

#### Before:
before

#### After:
after

### Alternate Designs

None were considered.

### Benefits

Syntax highlighting is now more accurate, all tests pass

### Possible Drawbacks

None

### Applicable Issues
#135

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1023,7 +1023,7 @@
         'include': '#generics'
       }
       {
-        'begin': '\\b(?:\\w*\\s*(\\.)\\s*)*([A-Z]\\w*)\\s*(?=\\[)'
+        'begin': '\\b(?:[A-Za-z]\\w*\\s*(\\.)\\s*)*([A-Z]\\w*)\\s*(?=\\[)'
         'beginCaptures':
           '1':
             'name': 'punctuation.separator.period.java'
@@ -1041,12 +1041,19 @@
       }
       {
         # Match possible generics first, eg Wow<String> instead of just Wow
-        'begin': '\\b((?:[A-Za-z]\\w*\\s*(\\.)\\s*)*[A-Z]\\w*)\\s*(?=<)'
+        'begin': '\\b((?:[A-Za-z]\\w*\\s*\\.\\s*)*[A-Z]\\w*)\\s*(?=<)'
         'beginCaptures':
           '1':
-            'name': 'storage.type.java'
-          '2':
-            'name': 'punctuation.separator.period.java'
+            'patterns': [
+              {
+                'match': '[A-Za-z]\\w*'
+                'name': 'storage.type.java'
+              }
+              {
+                'match': '\\.'
+                'name': 'punctuation.separator.period.java'
+              }
+            ]
         # Stop after a successful generic match or if we are not at a terminator
         'end': '(?<=>)|(?!;)'
         'patterns': [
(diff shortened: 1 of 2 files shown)


spec/java-spec.coffee
```diff
@@ -935,6 +935,21 @@ describe 'Java grammar', ->
...
 
+  it 'tokenizes fully qualified types', ->
+    lines = grammar.tokenizeLines '''
+      class A {
+        java.util.Set<java.util.List<K>> varA = null;
+        java.lang.String varB = null;
+      }
+    '''
+
+    expect(lines[1][1]).toEqual value: 'java.util', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.definition.variable.java', 'storage.type.java']
```
sadikovi: Not sure about that. I remember trying to fix it and eventually decided to look for another approach, so we can separate each package token and add scope for a `.`.

grammars/java.cson
```diff
@@ -1023,7 +1023,7 @@
         'include': '#generics'
       }
       {
-        'begin': '\\b(?:[A-Z]\\w*\\s*(\\.)\\s*)*([A-Z]\\w*)\\s*(?=\\[)'
```
sadikovi: Why do we need to remove `A-Z`? Should we do similar to other statements `A-Za-z`?

spec/java-spec.coffee
```diff
@@ -935,6 +935,21 @@ describe 'Java grammar', ->
...
+    expect(lines[1][2]).toEqual value: '.', scopes: [ 'source.java', 'meta.class.java', 'meta.class.body.java', 'meta.definition.variable.java', 'storage.type.java', 'punctuation.separator.period.java' ]
```
sadikovi: No spaces after `[` and before `]`.

spec/java-spec.coffee
sadikovi: Have you tried any other approaches? Interesting to see if there is a way of doing so. If not or it is too complicated, then we merge this PR and think of improving it later.

brennenberkley: Are you saying it would be better if the periods had a scope?

brennenberkley:
This new commit does make one test fail:
```
class Test {
        private Test.Double something;
      }
```
It wants ```Test.Double``` to be ```storage.type.java```, but with this change it now recognizes ```Test``` and ```Double``` as ```storage.type.java``` individually. This seems better to me, should I change that test? With the default theme it does make the periods a different color now:
screen shot 2018-07-18 at 2 06 37 am

spec/java-spec.coffee
```diff
@@ -935,6 +935,25 @@ describe 'Java grammar', ->
     expect(lines[12][3]).toEqual value: '123illegal', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.body.java', 'invalid.illegal.identifier.java']
     expect(lines[12][4]).toEqual value: ';', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.body.java', 'punctuation.terminator.java']
 
+  it 'tokenizes fully qualified types', ->
+    lines = grammar.tokenizeLines '''
+      class A {
+        java.util.Set<java.util.List<K>> varA = null;
+        java.lang.String varB = null;
+      }
```
sadikovi: Can you also add test for something like this?
```java
java.util.List<Integer> a = new java.util.ArrayList<Integer>();
java.util.List<Integer> b = abc.def.ghi();
```

And also check generic in the test.

sadikovi: That is the point, it should not be highlighted as storage type.

sadikovi: But no need to add it, I think there are tests that check function calls.

sadikovi:
Great job! I found that `java.lang.String[] arr;` does not seem to be scoped correctly.

brennenberkley:
I should have caught that, it's right above the other stuff I changed. New commit fixes it and adds a test

grammars/java.cson
```diff
@@ -1023,10 +1023,19 @@
...
+                'match': '[A-Za-z]\\w*'
+                'name': 'storage.type.java'
+              }
+              {
+                'match': '\\.'
+                'name': 'punctuation.separator.period.java'
+              }
+            ]
           '2':
             'name': 'storage.type.object.array.java'
```
sadikovi: I am not sure if this should apply to the last part of the class name or all parts, i.e. for `java.lang.String`, either String only has `storage.type.object.array.java`, or `java`, `lang` and `String` have it.

sadikovi:
@50Wliu I am going to merge this PR? Do you have any review comments? Thanks!

sadikovi:
Merged. Thank you @brennenberkley!

PR #148, opened 2018-07-24
Fix scope of generics within parenthesis

### LINKED ISSUE #132: Generic angle brackets have the wrong scope in situations like variable = (A<B>) value.
### Description

This is an issue reported as part of #127.

Tested on the latest master branch 7a42895fead16e2c39c6667fc9e7ce0f7f922ac4.
Code to reproduce:
```java
class A {
  void func() {
    B<C> value = null;
    value = (B<C>) obj;
  }
}
````

The angle brackets on the line `value = (B<C>) obj` have the wrong scope `keyword.operator.comparison.java`, but should be `punctuation.bracket.angle.java`.

Screenshot (click on the image for higher resolution):

screenshot from 2018-02-21 09-50-28

### Steps to Reproduce

1. Open Atom new tab
2. Select Java language
3. Paste the code sample above
4. Inspect angle brackets DOM elements in DevTools console

Angle brackets in the code snippet above are scoped as `punctuation.bracket.angle.java`

Angle brackets in the example above are scoped as `keyword.operator.comparison.java`. 

Reproducible 100% - looks like a bug I might have introduced when fixing generics highlighting related to comparison operators.

### Versions
Tested on both OS X and Ubuntu. 
Atom 1.23.3+.
`language-java` is the latest master.

### Additional Information
This was reported originally in #127. For more information, please see that issue.

### PR DESCRIPTION
**Pull request by brennenberkley**
The angle brackets in ```value = (B<C>) obj;``` were being treated as comparison operators instead of punctuation.

### Description of the Change

Generics within parenthesis were not being recognized, so I added a pattern to match them before anything else.

### Alternate Designs

None

### Benefits

Syntax highlighting will be more accurate, all tests pass

### Possible Drawbacks

none

### Applicable Issues

Fixes #132

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1152,6 +1152,14 @@
             'name': 'punctuation.bracket.round.java'
         'patterns': [
           {
+            'match': '<\\s*(?:[A-Za-z]\\w*\\s*\\.\\s*)*[A-Za-z]\\w*\\s*>'
+            'captures':
+              '0':
+                'patterns': [
+                  'include': '#generics'
+                ]
+          }
+          {
             'include': '#code'
           }
         ]
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -1177,6 +1177,29 @@ describe 'Java grammar', ->
     expect(lines[2][13]).toEqual value: 'null', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'constant.language.java']
     expect(lines[2][14]).toEqual value: ';', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'punctuation.terminator.java']
 
+  it 'tokenizes generics within parenthesis', ->
+    {tokens} = grammar.tokenizeLine 'A<B> value = (A<B>) obj;'
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by brennenberkley** [comment]
Before:
screen shot 2018-07-24 at 10 32 19 am

After:
screen shot 2018-07-24 at 10 32 01 am

**Comment by sadikovi** [comment]
Thanks for opening a PR. 

I have a couple of minor concerns with this approach; plus, it does not seem to work on the following cases:
```java
value = (B<T<C>>) a;
value = (B<T<? extends C>>) a;
```

And potentially other weird generic combinations. Did you look to see if there is anything we do already for it?

brennenberkley:
I didn’t see anything else that that handles generics within parenthesis, only generics that are directly in classes and functions. 

Do you have any suggestions for a better approach?

sadikovi:
Your approach seems reasonable, we just need to make it work for a wider set of generic expressions:). I will also look into the code closely and will let you know if there is anything we are missing in this approach.

Well done with the initial fix!

sadikovi:
@brennenberkley any progress on this issue? I was thinking that we could merge your fix for now and open a follow-up issue. Let me know if you are happy with this approach. I will also try poking around a bit more and see if there is anything else we can do.

brennenberkley:
I have not worked on it recently. We can pull it for now

PR #149, opened 2018-07-26
Simplify some syntax that was more complex than necessary

brennenberkley:
### Description of the Change

The end tag for the block was ```(?<=>)|(?!;)```, which could be simplified to ```(?!;)``` since that always matches the next character after the begin block. This makes the size of the block 0, so nothing is actually happening inside the block, just in the beginCaptures, so I just converted the beginCaptures to their own matching group.

### Alternate Designs

I tried to figure out if this was originally intended to do something else, but I couldn't get that syntax to do anything useful even in simple cases. 

### Benefits

The code is easier to understand.

### Possible Drawbacks

This might have had some other purpose I'm not seeing, but if so it wasn't working. This fix doesn't change the behavior at all. 

### Applicable Issues

None

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1049,9 +1049,9 @@
         ]
       }
       {
-        # Match possible generics first, eg Wow<String> instead of just Wow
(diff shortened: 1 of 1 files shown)


sadikovi:
Yes, you are right, we were already including generics there.

sadikovi:
Merged, thanks!

PR #150, opened 2018-07-26
Alphabetize the repository section

brennenberkley:
It was mostly in order already, but it looks like a few sections were added in the wrong places.

### Benefits

It makes the code easier to navigate. The order of the repository does not matter, so nothing will be broken.


sadikovi:
Good point, will merge.

PR #161, opened 2018-09-15
Fix regression of field values being incorrectly highlighted

### LINKED ISSUE #158: Field values are incorrectly highlighted
_From @rickyah on September 3, 2018 8:55_

Issue Type: <b>Bug</b>

Using this snippet makes any text written after the BUG_SAMPLE variable to render as a string:

```java
final class Example
{
    private static final String MODULE = "test"

    private static final String BUG_EXAMPLE = "window." + TEST + "aFunction();";

    Example(PopupWebView popupWebView)
    {
        this.popupWebView = popupWebView;
    }
}
```

Atom (language-java 0.30.0) highlights this correctly, but VSCode (on master) ends the variable declaration with the `=`.

screenshot from 2018-09-12 15-56-25
screenshot from 2018-09-12 15-56-50

sadikovi:
### Requirements

### Description of the Change

This PR fixes the regression reported in #158. This happened because our variable definition now ends with `=`, and the rest is supposed to be matched by the code, which does happen correctly, but it breaks in certain scenarios when using class fields, e.g.
```java
class A {
  private String b = a + "a()" + a() + "" + "a();" + "" + a() + abc + b() + "b();";
  private String c = "a / a();";
  private int d[] = a + "a();" + func();

  int abcd() {
    return 1;
  }
}
```

### Before
before

### After
after

All of the `b`, `c`, and `d` would break a subsequent highlighting and scope of method `abcd`.

The fix is fairly straightforward and might seem as a hack. We define `#class-fields-and-methods` scope instead of `#methods` and first match anything like this `... = [...];`, where `[...]` is covered by `#code`, and only then we match methods. This seems to work quite well.

I also fixed comments in methods, while I was there. Added tests for both methods and class fields.

### Alternate Designs

I considered reverting #146 at some point, but then we would still have to solve `;` issue in try with resources, and, to be fair, the change in that PR was more or less reasonable.

I also tried updating pattern for `#methods` scope trying to restrict it, so that it does not show up in the expression part. It turned out to be quite difficult, mostly because of the return type and parameter types being potentially anything, which complicated the task. For example, I tried writing something like this: 
```cson
'methods':
    'begin': '(?!new)(?=[\\w<].*\\s+)(?=[\\w\\s\\[\\]<>,\\.\\?/\\*]+\\()'
```
But it did not work in certain cases.

### Benefits

Fixes the regression introduced by #146.

### Possible Drawbacks

Could potentially miss coverage of some other case.

### Applicable Issues

Closes #158

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -438,9 +438,6 @@
         'name': 'comment.block.empty.java'
       }
       {
-        'include': 'text.html.javadoc'
-      }
-      {
         'include': '#comments-inline'
       }
     ]
@@ -771,7 +768,7 @@
       }
     ]
   'methods':
-    'begin': '(?!new)(?=[\\w<].*\\s+)(?=([^=/]|/(?!/))+\\()'
+    'begin': '(?!new)(?=[\\w<].*\\s+)(?=[\\w\\s\\[\\]<>,\\.\\?/\\*]+\\()'
     'end': '(})|(?=;)'
     'endCaptures':
       '1':
@@ -780,6 +777,9 @@
     'patterns': [
       {
         'include': '#storage-modifiers'
+      }
+      {
+        'include': '#comments'
       }
       {
         'begin': '(\\w+)\\s*(\\()'
@@ -801,7 +801,7 @@
             'include': '#parens'
           }
           {
-            'include': '#comments-inline'
+            'include': '#comments'
           }
         ]
       }
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -352,9 +352,9 @@
     ]
   'class-fields-and-methods':
     # Pattern to match both class fields and method declarations inside class body.
-    # For class fields, we only need to match part after '=', because the first part will be matched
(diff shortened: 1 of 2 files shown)


grammars/java.cson
```diff
@@ -437,9 +456,6 @@
         'match': '/\\*\\*/'
         'name': 'comment.block.empty.java'
       }
-      {
-        'include': 'text.html.javadoc'
```
sadikovi: We do not have `text.html.javadoc` scope, instead we added `comments-javadoc`, and it is normally added separately.

grammars/java.cson
```diff
@@ -801,7 +817,7 @@
             'include': '#parens'
           }
           {
-            'include': '#comments-inline'
+            'include': '#comments'
```
sadikovi: `#comments` includes inline ones.

sadikovi:
@50Wliu could you review this PR? I updated the description on why I did it this way. Thanks!

This should unblock the `0.31.0` release.

sadikovi:
ping @50Wliu

PR #163, opened 2018-09-23
Variable highlighting in for-each loop

sadikovi:
### Requirements

### Description of the Change

I found when working on other issue that variables in for-each loops are not scoped correctly, for example:
```java
class Test {
  void func() {
    // correctly highlighted in for loop
    for (int i = 0; i < 10; i++) {
      // do something
    }

    // incorrectly highlighted and scoped in for-each loop
    for (int i : elements) {
      // do something
    }

    // incorrectly highlighted and scoped in for-each loop
    for (HashMap<String, String> map : elementsFunc()) {
      // do something
    }
  }
}
```

Before patch:
before

After patch:
after

I simply extended the list of characters that terminate variable patterns and added `:` as such. Added test to cover this scenario.

### Alternate Designs

None were considered.

### Benefits

Fixes variable highlighting in for-each loops.

### Possible Drawbacks

Should not be any.

### Applicable Issues

None

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1356,14 +1356,14 @@
         \\s+
         [A-Za-z_$][\\w$]* # At least one identifier after space
         ([\\w\\[\\],$][\\w\\[\\],\\s]*)? # possibly primitive array or additional identifiers
-        \\s*(=|;)
+        \\s*(=|:|;)
       )
     '''
-    'end': '(?=\\=|;)'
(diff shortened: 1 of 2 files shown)


sadikovi:
@50Wliu could you review this PR? Thanks!

I found this inconsistency when working on Java 10 var support, so I thought I might as well fix it (more or less).

sadikovi:
ping @50Wliu

PR #164, opened 2018-09-23
Java 10 local variable "var" type support

### LINKED ISSUE #162: Highlight `var` keyword
Java 10 `var` keyword is not highlighted, which makes declarations visually ambiguous with calls. That makes code significantly harder to navigate.

sadikovi:
### Requirements

### Description of the Change

This PR adds support for `var` storage type that is used in Java 10 to indicate local variable with inferred type. Details can be found here: http://openjdk.java.net/jeps/286.

I created a new pattern `variables-local` specifically for declarations `var x = ...`. This allows to only highlight `var` keyword when it is actually a storage type. I tested on the following code:

```java
class Test {
  void func() {
    A a = new A();
    var b = new B();
    // should be scoped as a variable name
    C var = new C();
    // should be scoped as a variable name
    var = new C();
    {var d = new D();}

    for (var d : elements()) {
      // should highlight as storage type
    }
  }
}
```

Before:
before

After:
after

I added `#variables-local` to `#code` just next to `#variables` pattern. This has it's benefits, such as working out of the box inside for loops and code blocks. Unfortunately, this highlights `var x = ...` when defined in class body, which will result in compilation error based on the JDK doc, but I guess it should be okay.

### Alternate Designs

None were considered apart from adding `variables-local` pattern separately to scopes where it should be defined, but I think this would have been messy and would have still had issues.

### Benefits

Adds a feature of `var` support.

### Possible Drawbacks

Should not impact any existing code, even if they use `var` as a variable name.

### Applicable Issues

Closes #162

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,15 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an issue is closed for lack of response
+daysUntilClose: 28
+
+# Label requiring a response
+responseRequiredLabel: more-information-needed
+
+# Comment to post when closing an issue for lack of response. Set to `false` to disable.
+closeComment: >
+  This issue has been automatically closed because there has been no response
+  to our request for more information from the original author. With only the
+  information that is currently in the issue, we don't have enough information
+  to take action. Please reach out if you have or find the answers we need so
+  that we can investigate further.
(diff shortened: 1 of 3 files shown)


sadikovi:
@50Wliu would you mind reviewing this PR? This is how I see it being implemented, I am open to suggestions and overall feedback on having this feature. Thanks!

sadikovi:
ping @50Wliu

PR #175, opened 2018-11-23
Fix enums with override methods

### LINKED ISSUE #173: Overrided methods in enum is not highlighted properly
### Prerequisites

### Description

Overrided methods in enum is not highlighted properly.

```java
enum TYPES {

    TYPE_A {
        @Override
        int f() {
            return 1;
        }
    }
    , TYPE_B {
        @Override
        int f() {
            return 2;
        }
    }, TYPE_DEFAULT;

    public int f() {
        return 0;
    }
}
```
image

### Steps to Reproduce

TYPE_A, TYPE_B and TYPE_DEFAULT should have the same scopes.

the three definitions of `f` should have proper highlight.

### Versions

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

sadikovi:
### Requirements

### Description of the Change

This PR extends support for enums and adds a pattern to handle enum's constant fields with method overrides. It is allowed to have such fields mixed with the previous two types that we have added.

The fix is fairly simple - adding another pattern to handle the case of enum field starting with `{` and ending with `}` similar to inner class. For that reason, I used `include: #class-body` instead of `include: #code`.

### Alternate Designs

None were considered, since it is fairly simple fix.

### Benefits

Fixes enum highlighting for enums with override methods.

### Possible Drawbacks

None. It only affects changes for enums.

### Applicable Issues

Closes #173

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -587,6 +587,26 @@
               {
                 'include': '#comments'
               }
+              # Represents enum's constant field with the code overrides
+              {
+                'begin': '(\\w+)\\s*({)'
+                'beginCaptures':
+                  '1':
+                    'name': 'constant.other.enum.java'
+                  '2':
+                    'name': 'punctuation.bracket.curly.java'
+                'end': '\\}'
+                'endCaptures':
+                  '0':
+                    'name': 'punctuation.bracket.curly.java'
+                'patterns': [
+                  {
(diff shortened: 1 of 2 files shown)


sadikovi:
ping @50Wliu

sadikovi:
I think I am going to merge this PR. It is very low risk (only affects enums) and it is unit tested.

PR #188, opened 2019-01-09
Add support for `final` in `catch` parameters

0xc0:
### Description of the Change

The regex for the `catch` parameters currently does not support `final`, meaning that `final` is highlighted as a type. This change adds an optional group to account for the possibility of the Java `final` modifier.

### Alternate Designs

None, as prepending the existing regex is a simple change.

### Benefits

Correct syntax highlighting when `final` is within `catch` parameters

### Possible Drawbacks

None.

### Applicable Issues

None.

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1330,11 +1330,13 @@
                 'name': 'punctuation.catch.separator.java'
               }
               {
-                'match': '([a-zA-Z$_][\\.a-zA-Z0-9$_]*)\\s*(\\w+)?'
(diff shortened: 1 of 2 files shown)


sadikovi:
Are there any other keywords available for the catch clause? Have you considered an option of including `#storage-modifiers`? Can you add a unit test?

0xc0:
@sadikovi Updated to use `#storage-modifiers` (better idea) and added a unit test

PR #190, opened 2019-01-12
Add extends and implements to enums scope

### LINKED ISSUE #185: Enum implements does not highlight
### Description

image

```java
class A {
    public enum _Fields implements org.apache.thrift.TFieldIdEnum {

        WHY((short) 1, "why");

        static {
            for (_Fields field : EnumSet.allOf(_Fields.class)) {
                byName.put(field.getFieldName(), field);
            }
        }

        /**
         * Find the _Fields constant that matches fieldId, or null if its not found.
         */
        public static _Fields findByThriftId(int fieldId) {
            switch(fieldId) {
                case // WHY
                1:
                    return WHY;
                default:
                    return null;
            }
        }
    }
}
```

version 7c3d0d3e37920790cd3afb4e68830bdb215851a3

sadikovi:
### Requirements

### Description of the Change

This PR adds `extends` and `implements` to `enums` scope. Currently we have enum scope separate from class and interface, because of the complexity of enum's representations, though we did not migrate `extends` and `implements` to the enum. This work fixes the issue by copying those blocks of code from class scope and adding them to enum scope with some minor changes.

### Alternate Designs

None were considered.

### Benefits

Fixes a regression for enum's `extends` and `implements`.

### Possible Drawbacks

This patch introduces a bit of code duplication between class and enum, mostly because of the separation of those scopes.

### Applicable Issues

Closes #185

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -571,6 +571,38 @@
     'name': 'meta.enum.java'
     'patterns': [
       {
+        'begin': '\\b(extends)\\b'
+        'beginCaptures':
+          '1':
+            'name': 'storage.modifier.extends.java'
+        'end': '(?={|\\bimplements\\b)'
+        'name': 'meta.definition.class.inherited.classes.java'
+        'patterns': [
+          {
+            'include': '#object-types-inherited'
+          }
+          {
+            'include': '#comments'
+          }
+        ]
+      }
+      {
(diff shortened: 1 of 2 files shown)


sadikovi:
This is fairly straightforward change. I am not aware of any way of making it better without even bigger refactoring.

PR #200, opened 2019-06-14
Fix underscore in class names

### LINKED ISSUE #198: Underscore "_" not recognized inside type qualifier
### Description

When a type qualifier starts with "_", it breaks the grammar for the rest of the line.

image

### Fix

I replaced all `[A-Za-z]\\w*` to `[a-zA-Z$_][\\.a-zA-Z0-9$_]*`, and `(\\w+\\.)*[A-Z]+\\w*` to `([a-zA-Z$_][a-zA-Z$_0-9]*\\.)*[A-Z]+\\w*)` it seems fixed.

sadikovi:
### Requirements

### Description of the Change

I updated patterns that relate to storage types to include underscore wherever it was missing; also updated variable pattern to capture storage types with underscore.

Added unit test.

### Alternate Designs

No alternative designs were considered.

### Benefits

Fixes issue of having package and class names that start with underscore.

### Possible Drawbacks

I had to comment out line 832 in `java-spec.coffee`, "_1" value is now highlighted as `storage.type.java`. I will open issue to fix this later. Should not affect anything else, as far as I am concerned.

### Applicable Issues

Fixes #198

### DIFF
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -828,10 +828,6 @@ describe 'Java grammar', ->
         {tokens} = grammar.tokenizeLine '1_'
         expect(tokens[0]).toEqual value: '1_', scopes: ['source.java']
 
-        # TODO: Fix this test, currently it has a scope of "storage.type.java", because it starts with underscore.
-        # {tokens} = grammar.tokenizeLine '_1'
-        # expect(tokens[0]).toEqual value: '_1', scopes: ['source.java']
-
         {tokens} = grammar.tokenizeLine '2639724263Q'
         expect(tokens[0]).toEqual value: '2639724263Q', scopes: ['source.java']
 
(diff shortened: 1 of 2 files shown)


spec/java-spec.coffee
```diff
@@ -828,8 +828,9 @@ describe 'Java grammar', ->
         {tokens} = grammar.tokenizeLine '1_'
         expect(tokens[0]).toEqual value: '1_', scopes: ['source.java']
 
-        {tokens} = grammar.tokenizeLine '_1'
-        expect(tokens[0]).toEqual value: '_1', scopes: ['source.java']
+        # TODO: Fix this test, currently it has a scope of "storage.type.java", because it starts with underscore.
```
50Wliu: This is probably a valid type anyway in Java, right?

sadikovi: Yes, it could be.

sadikovi: Let me check quickly.

Arcanemagus: It at least appears to be a [valid Identifier](https://docs.oracle.com/javase/specs/jls/se12/html/jls-3.html#jls-3.8).

sadikovi: Yes, I checked, it is a valid class name. I think I am going to remove the test.

PR #206, opened 2019-09-12
Fix Javadoc @link tag highlighting broken issue

CsCherrYY:
### Requirements

### Description of the Change

This PR uses a non-greedy match instead of a greedy match in the highlighting. This change may bring correct highlighting in the Javadoc @link tag when more than one @link tag in the same line.

### Alternate Designs

 It is a simple fix so no alternate designs were considered.

### Benefits

Fix Javadoc @link tag highlighting broken problem.

### Possible Drawbacks

None.

### Applicable Issues

[#958](https://github.com/redhat-developer/vscode-java/issues/958)

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -525,7 +525,7 @@
               'name': 'entity.name.type.class.java'
         }
         {
-          'match': '{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]*\\)).*}'
+          'match': '{(@link)\\s+(\\S+)?#([\\w$]+\\s*\\([^\\(\\)]*\\)).*?}'
           'captures':
             '1':
               'name': 'keyword.other.documentation.javadoc.java'


sadikovi:
Your PR breaks the test, can you have a look why it fails?

CsCherrYY:
> Your PR breaks the test for `link`, can you have a look why it fails?

OK. But the fix seems to work well in VS Code. Could you tell me how can I do the tests myself?

sadikovi:
You can run `apm test` to run the tests in the repository. You can also have a look at Atom Flight Manual, for example, https://flight-manual.atom.io/hacking-atom/sections/writing-specs/.

I would suggest adding a new test for your changes with multiple link tags in the javadoc.

CsCherrYY:
> You can run `apm test` to run the tests in the repository. You can also have a look at Atom Flight Manual, for example, https://flight-manual.atom.io/hacking-atom/sections/writing-specs/.
> 
> I would suggest adding a new test for your changes with multiple link tags in the javadoc.

I think it is inappropriate in the original test of java-spec.coffee in line 2553, `label {@link Class#method()}}` , so I deleted the extra brace and add the multiple @link tags test.

Issue #183, opened 2018-12-26
$ recognized as word boundary


felixhao28:
### Description

```java
class $Apple {
    void $func$tion() {
        this_x = 5;
        this$xx = 4;
        abc$xx = 5;
        x = 5;
        int abc$xx = 5;
        int abc_xx = abc$xx;
        $App$le this$_xx = abc$xx;
    }
}
```

Code like this messes up the highlight pretty bad. This is what looks like now:

![image](https://user-images.githubusercontent.com/2417391/50443727-9f32b180-093f-11e9-80e2-369be788f29f.png)

This is what **should look like**:

![image](https://user-images.githubusercontent.com/2417391/50443563-e3718200-093e-11e9-91f8-b8970db6b26d.png)

Version: 
7c3d0d3e37920790cd3afb4e68830bdb215851a3

sadikovi:
Other issues related to $: #33.

PR #208, opened 2019-11-21
Parse class identifier containing a dollar char

Eskibear:
### Description of the Change

`\w` equals to`A-Za-z0-9_`.  `$` is not contained, which is allowed to be part of the class identifier. 

### Alternate Designs

None were considered.

### Benefits

Can highlight class name containing `$`.

### Possible Drawbacks

I don't see any. 

### Applicable Issues

https://github.com/redhat-developer/vscode-java/issues/299
Screen Shot 2019-11-21 at 3 09 36 PM

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -243,7 +243,7 @@
       }
     ]
   'class':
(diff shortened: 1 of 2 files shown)


sadikovi:
Partially fixes #33 and #183

sadikovi:
Merged, thanks.

PR #210, opened 2019-11-24
Comments should not end 'keyword.control.new.java'

### LINKED ISSUE #209: Comments before inner class's class body break out the current scope
### Prerequisites

    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

####  Original

```java
IJobChangeListener listener = new JobChangeAdapter()
{
	@Override
	public void scheduled(IJobChangeEvent event) {
		if (event.getJob() instanceof UpdateClasspathJob) {
			jobInvocations[0] = jobInvocations[0] + 1;
		}
	}
};
```

image

Scope of `{`: 
* `punctuation.section.inner-class.begin.bracket.curly.java` 
* `meta.inner-class.java`

Scope of `event`:
* `variable.parameter.java`
* `meta.method.identifier.java`
* `meta.method.java`

#### Add a single line comment before class body

image

Scope of `{`: 
* `punctuation.section.block.begin.bracket.curly.java` 
* `meta.method.body.java`

Scope of `event`:
* `meta.function-call.java`
* `meta.method.body.java`
* `meta.method.java`

#### Add a multi line comment before class body

image

Scope is the same as single line comment.

### Steps to Reproduce

**Expected behavior:** The existence of comment does not affect the syntax of inner-class.

**Actual behavior:** Comments before inner class's class body break out the current scope

Vigilans:
### Description of the Change

Currently, the divide operator ('/') will end 'keyword.control.new.java' scope. But ('/') could also be the beginning of a comment, and it should not end the scope.

### Alternate Designs

No alternative considered.

### Benefits

Comment would not affect the highlighting of inner class, as described in #209.

### Possible Drawbacks

Nothing as far as I know.

### Applicable Issues

Fix #209

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -191,7 +191,7 @@
     'beginCaptures':
       '0':
         'name': 'keyword.control.new.java'
-    'end': '(?=;|\\)|,|:|}|\\+|\\-|\\*|\\/|%|!|&|\\||=)'
+    'end': '(?=;|\\)|,|:|}|\\+|\\-|\\*|\\/(?!\\/|\\*)|%|!|&|\\||=)'
     'patterns': [
       {
         'include': '#comments'
--- a/spec/java-spec.coffee
+++ b/spec/java-spec.coffee
@@ -1682,6 +1682,21 @@ describe 'Java grammar', ->
     expect(tokens[6]).toEqual value: 'Point', scopes: ['source.java', 'meta.function-call.java', 'entity.name.function.java']
     expect(tokens[9]).toEqual value: ';', scopes: ['source.java', 'punctuation.terminator.java']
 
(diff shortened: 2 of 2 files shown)


sadikovi:
Can you add a test case for this change?

Vigilans:
@sadikovi Unit test is added.

PR #212, opened 2019-11-25
More robust enum field's pattern logic

### LINKED ISSUE #211: Enum not correctly hightlighted when mixed with constructor and code overrides
### Description

Originally created in https://github.com/redhat-developer/vscode-java/issues/974 by @orblazer :

> Hello,
> The color of first item when the enum overwrite is wrong
> 
> ##### Environment
> * Operating System: Windows 10
> * JDK version: jdk1.8.0_212
> * Visual Studio Code version: 1.35.1
> * Java extension version: 0.46.0
> 
> ##### Steps To Reproduce
> 1. Create an enum with function
> 2. Override that function in enum items
> 
> [attach a sample project reproducing the error]
> [attach logs](https://github.com/redhat-developer/vscode-java/wiki/Troubleshooting#enable-logging)
> 
> ##### Current Result
> wrong
> 
> ##### Expected Result
> right
> 
> ##### Additional Informations
> The path for color is different :
> 
> **Path of current result :**
> wrong-dt
> 
> **Path of expected result :**
> right-dt
> 
> I think the path for the `current result` is right bug its just parsing is broken.

Vigilans:
### Description of the Change

The original logic uses three different patterns to match an enum constant field. This PR replaces them with a two-scope uniform pattern.

### Alternate Designs

To solve #211, an alternative would be to add another to match (constructor + code overrides), but it introduces too much duplicate code, and it still cannot highlight correctly when '{' is placed at new line.

### Benefits

* Can correctly highlight with all possible combinations of constructor & code overrides
* Can correctly highlight when '{' is placed at new line or with a comment in between
* Patterns are more neat and uniform

### Possible Drawbacks

* Undetected possible regression issues

### Applicable Issues

Fix #211 
Fix redhat-developer/vscode-java#974

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -619,8 +619,7 @@
               {
                 'include': '#comments'
               }
-              # Represents enum's constant field entry
-              {
+              { # Represents enum's constant field entry
                 'begin': '\\b(\\w+)\\b'
                 'beginCaptures':
                   '1':
@@ -636,7 +635,20 @@
                   {
                     'include': '#comments'
                   }
-                  { # Represents enum's constant field with the code overrides
+                  { # Represents enum's constructor, e.g. `("1", 123)` in `Test("1", 123)`
+                    'begin': '\\(',
+                    'beginCaptures':
+                      '0':
+                        'name': 'punctuation.bracket.round.java'
+                    'end': '\\)'
+                    'endCaptures':
+                      '0':
+                        'name': 'punctuation.bracket.round.java'
+                    'patterns': [
+                      'include': '#code'
+                    ]
+                  }
+                  { # Represents enum's body with code overrides
                     'begin': '{',
                     'beginCaptures':
                       '0':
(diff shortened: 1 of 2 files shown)


grammars/java.cson
```diff
@@ -619,49 +619,53 @@
...
+                    'endCaptures':
+                      '0':
+                        'name': 'punctuation.bracket.curly.java'
+                    'patterns': [
+                      # Use #class-body, because the enum field overrides enum class body.
+                      # It is similar to inner class.
+                      'include': '#class-body'
+                    ]
+                  }
+                  { # Represents enum's constant field with constructor, e.g. `Test(123)`
```
sadikovi: Can you move this block so it is before "code overrides block"? Also, comment is no longer relevant, let's just say this is "# Represents enum's constructor" or something similar.

grammars/java.cson
```diff
@@ -619,49 +619,53 @@
...
-                'end': '\\)'
-                'endCaptures':
-                  '0':
-                    'name': 'punctuation.bracket.round.java'
-                'patterns': [
                   {
-                    'include': '#code'
+                    'include': '#comments'
+                  }
+                  { # Represents enum's constant field with the code overrides
```
sadikovi: Comment is no longer relevant, should be "# Represents enum's body with code overrides" or something similar.

Vigilans:
New commit added to resolve the review conversations.

PR #214, opened 2019-11-27
Provide additional stop conditions for 'anonymous-classes-and-new' pattern

### LINKED ISSUE #179: `new` keyword messes with ternary operator when used in condition
### Description

```java
public class A {
    void f() {
        String sex = new Random().nextInt(2) == 0 ? "female" : "male";
        String sex = new Random().nextInt(2) ? "female" : "male";
        String sex = Random().nextInt(2) ? "female" : "male";
        String sex = new Random() ? "female" : "male";
    }
}
```
image

The first string in "true branch" should be recognized as `string.quoted.double.java`, like in the third line.

### LINKED ISSUE #180: Function call + array accessor + new operator introduces extra meta.function-call.java scopes
### Description

```java
public class A {
    void f() {
        int a = 1;
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        g(education[new Random()]);
        int a = 1;
    }

    void g(Object o) {
        int a = 1;
    }
}
```

image

As you can see, although the highlight is still correct, there are more `meta.function-call.java` than we need.

### LINKED ISSUE #192: Minus operator confused with variable after new constructor
Current behavior

image

Expected:

image

Code:

```java
class A {
    void f() {
        int a = new Date().getTime() - start.getTime();
    }
}
```

Fix:

Change [this line](https://github.com/atom/language-java/blob/master/grammars/java.cson#L194) from 

```
'end': '(?=;|\\)|,|:|}|\\+)'
```

 to

```
'end': '(?=;|\\)|,|:|}|\\+|\\-)'
                           ^^^
```

### LINKED ISSUE #213: Inconsistency of variable and new-variable method call
### Description

Consider following code:
```java
// variable method call
Date date = new Date();
date.toString();

// new-variable method call
new Date().toString();
```
The former is a method call, and the color of dot is `#abb2bf`:

The latter is a function call (toString), where the dot has no meaning, and the color of dot is `#e06c75`:

**Expected behavior:** (xxx.method) should all be method call no matter how the varible is derived

**Actual behavior:** The meaning of dot method call is inconsistent

Vigilans:
### Description of the Change

* Dot should also end  'anonymous-classes-and-new', in order to propery highlight method call of temporary variable created by 'new'.

* Question mark should also end 'anonymous-classes-and-new', in order to proper break out circumstances like `new Random() ? true : false`;

* Right square brace should also end 'anonymous-classes-and-new', for circumstances like `[new Random()]`.

* Xor should also end `anonymous-classes-and-new`, for circumstances like `new Integer(1) ^ new Integer(2)`.

### Alternate Designs

~~Non was considered.~~

Having to enumerate these much punctuations made me think of proposing a general and simple regex to end new expression. But soon I found out that I couldn't find a way to tell when `)` or `]` is met, whether there will be a `{}` block next to it.

### Benefits

* The meaning of dot method call by variable and new-variable is consistent.
* Every punctuation that could end a new expression is covered.

### Possible Drawbacks

No obvious drawback.

### Applicable Issues

Fix #213 -> dot
Fix #179 -> question mark
Fix #180 -> right square
Fix #192 -> xor


spec/java-spec.coffee
```diff
@@ -1755,6 +1768,41 @@ describe 'Java grammar', ->
     expect(lines[7][19]).toEqual value: 'start', scopes: expected
     expect(lines[8][19]).toEqual value: 'start', scopes: expected
     expect(lines[9][19]).toEqual value: 'start', scopes: expected
+    expect(lines[10][19]).toEqual value: 'start', scopes: expected
+
+  # See issue https://github.com/atom/language-java/issues/180
```
sadikovi: Formatting is off.

sadikovi: Same below.

spec/java-spec.coffee
```diff
@@ -1755,6 +1768,41 @@ describe 'Java grammar', ->
...
+
+    expect(lines[3][1]).toEqual value: 'g', scopes: expected
+    expect(lines[4][1]).toEqual value: 'g', scopes: expected
+    expect(lines[5][1]).toEqual value: 'g', scopes: expected
+    expect(lines[6][1]).toEqual value: 'g', scopes: expected
+    expect(lines[7][1]).toEqual value: 'g', scopes: expected
+    expect(lines[8][1]).toEqual value: 'g', scopes: expected
+    expect(lines[9][1]).toEqual value: 'g', scopes: expected
+    expect(lines[10][1]).toEqual value: 'g', scopes: expected
+    expect(lines[15][3]).toEqual value: 'a', scopes: ['source.java', 'meta.class.java', 'meta.class.body.java', 'meta.method.java', 'meta.method.body.java', 'meta.definition.variable.java', 'variable.other.definition.java']
```
sadikovi: Does this test fail without your change?

Vigilans: See #180, it will produce a nested scope like this:
```
variable.other.definition.java
meta.definition.variable.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.function-call.java
meta.method.body.java
meta.method.java
meta.class.body.java
meta.class.java
source.java
```

sadikovi: I am well aware of the issue. I was only asking if you verified that the test fails without your changes.

Vigilans: I see. I test it by removing `\\[` in the `end` regex, and it produces such failures:

image

sadikovi:
Please, enable edits from maintainers on your pull request.

sadikovi:
Merging to master, thanks!

PR #220, opened 2019-12-07
Fix catch parameter highlighting

### LINKED ISSUE #219: Catch parameter not correctly highlighted when declared in new line or with a comment in between
### Description

Catch parameter name is highlighted as type when declared in new line or with a comment in between.

### Steps to Reproduce

Paste the following code to editor:

```java
    try {
        // something
    } catch // this is a comment
    (NoSuchMethodException |
        SecurityException
      | InstantiationException | llegalAccessException |
      IllegalArgumentException | invocationTargetException 
      e
    ) {
        e.printStackTrace();
    }
}

    try {
        // something
    } catch // this is a comment
    (NoSuchMethodException |
        SecurityException
      | InstantiationException | llegalAccessException |
      IllegalArgumentException | invocationTargetException /* comment */ e
    ) {
        e.printStackTrace();
    }
}
```

**Expected behavior:** `e` is highlighted as `variable.parameter.java`

**Actual behavior:** 

`e` is highlighted as `storage.type.java`:
image

Vigilans:
### Description of the Change

Uses a two-scope begin/end model to correctly highlight catch parameter.

It wraps `Exception ... |` or `Exception ... )` as a scope, allowing parameter to only sit in this scope, to resolve the parsing sequence issue (first exception type, then parameter identifier).

#### A more complex two-scope design:
* scope 1: `(...Exception`
* scope 2: `|...Exception`
* scope 3: `parameter...)`

Problem: cannot correctly highlight when there's a storage-modifer (like final)

#### Recursive design:
```yaml
catch-parameters:
  begin: /(?<=\()|(\|)/
  beginCapture.0.name: 'catch.separator'
  end: /(?=\))/
  patterns:
    - include: storage-modifier
    - include: exception-type
    - include: catch-parameters # recursive
    - include: paramter-identifier
```

It makes use of the fact that `catch (ex1 | ex2 | ex3 param)` is in terms of structure the same as `catch (ex1 (ex2 (ex3 param)` (every time a `|` is met, the remaining part is actually a new catch-parameter pattern) 

Problem: still cannot elegantly handle the base case: `catch (final ex3 param)`. 

### Benefits

Can correctly highlight catch parameter when it is placed in new line or with a comment in between.

### Possible Drawbacks

Don't see any afak.

### Applicable Issues

Fix #219

### DIFF
--- a/grammars/java.cson
+++ b/grammars/java.cson
@@ -1367,16 +1367,25 @@
                 'include': '#storage-modifiers'
               }
               {
-                'match': '\\|'
-                'name': 'punctuation.catch.separator.java'
-              }
-              {
-                'match': '([a-zA-Z$_][\\.a-zA-Z0-9$_]*)\\s*(\\w+)?'
-                'captures':
-                  '1':
+                'begin': '[a-zA-Z$_][\\.a-zA-Z0-9$_]*'
+                'beginCaptures':
+                  '0':
                     'name': 'storage.type.java'
-                  '2':
-                    'name': 'variable.parameter.java'
+                'end': '(\\|)|(?=\\))'
+                'endCaptures':
+                  '1':
+                    'name': 'punctuation.catch.separator.java'
+                'patterns': [
+                  {
+                    'include': '#comments'
+                  }
+                  {
+                    'match': '\\w+'
+                    'captures':
+                      '0':
+                        'name': 'variable.parameter.java'
+                  }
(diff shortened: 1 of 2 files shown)


Vigilans:
I enriched the two testcases by inserting comments into possbile locations related with my changes.

sadikovi:
Merging to master, thanks!

PR #249, opened 2021-10-05
Install the exact version instead of range

sadick254:
`^version` means install any version that is compatible with the range.
This means npm can even install tree-sitter-javascript version 0.15
since in semver 0.15 should be compatible with 0.19. tree-sitter seems
to have broken semver. 0.15 is not compatible with 0.19

### DIFF
--- a/grammars/tree-sitter-java.cson
+++ b/grammars/tree-sitter-java.cson
@@ -1,7 +1,7 @@
 name: 'Java'
 scopeName: 'source.java'
 type: 'tree-sitter'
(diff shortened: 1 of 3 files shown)


sadikovi:
Thanks for opening a PR, I think this may not work since it requires my special tree-sitter-java-dev version which contains bug fixes. I tried to push them upstream to tree-sitter-java but those were rejected. Maybe you can fix it as
```
"tree-sitter-java-dev": "0.16.0-dev2"
```

